PR #1054, opened 2014-07-24
[Fix Bug 994927] Add map showing member locations to group page.

Sancus:
Fixed https://github.com/mozilla/mozillians/pull/1034


alex-mayorga:
¡Hola!

So where does one see this map?

I'm a vouched mozillian but I don't see it.

Am I missing something obvious?

¡Gracias!

glogiotatidis:
This code was never merged, that's why you can't find the map.

PR #1144, opened 2015-01-07
Multiple e-mail support

johngian:


johngian:
@akatsoulas r?

akatsoulas:
Personally I would organize the commits into undo-able chunks - commits that easy and meaningful  to revert. For example maybe it makes sense to be able to revert to the old and new functionality with just one commit, thus merging the first three. Either way it totally up to you how you are going to handle this.

mozillians/templates/phonebook/profile.html
```diff
@@ -261,6 +261,19 @@ <h1 class="p-name">
           </div>
         {% endif %}
 
+        {% if profile.alternate_emails.exists() %}
+          <div id="alternate_email" class="profile-entry">
+            <h3><i class="icon-envelope-o"></i> {{ _('Alternate email addresses') }}</h3>
```
akatsoulas: Without knowing what this class is about, I am not sure if -o is a suitable suffix for the class name.

johngian: This is used all around the project to render the envelope icon for emails.

mozillians/phonebook/urls.py
```diff
@@ -13,6 +13,12 @@
     url(r'^user/edit/$', 'views.edit_profile', name='profile_edit'),
     url(r'^u/(?P<username>[\w.@+-]+)/$', 'views.view_profile',
         name='profile_view'),
+    url(r'^user/edit/emails/$', 'views.edit_emails',
+        name='edit_emails'),
+    url(r'^user/delete/email/(?P<email_pk>\d+)$', 'views.delete_email',
```
akatsoulas: Append `/`

mozillians/phonebook/urls.py
```diff
@@ -13,6 +13,12 @@
...
+        name='delete_email'),
+    url(r'^user/primary/email/(?P<email_pk>\d+)$', 'views.change_primary_email',
```
akatsoulas: Append `/`

johngian:
Regarding commits I would prefer to have small atomic commits. It makes sense for me to either squash 71192ea7bce0e37ede958bf8f80f0ad32b2b2229 d318a04505346dbb9989aae60e34a1819bb725b3 or split them to even more commits.

mozillians/phonebook/views.py
```diff
@@ -232,6 +224,73 @@ def edit_profile(request):
...
+    alternate_email = alternate_emails.get(pk=email_pk)
+    primary_email = user.email
+
+    # Change primary email
+    user.email = alternate_email.identifier
+
+    # Turn primary email to alternate
+    alternate_email.identifier = primary_email
+
+    with transaction.commit_on_success():
```
akatsoulas: Why are you using this? I am not sure if you actually need to do this in a single transaction. Moreover, `commit_on_success` is now a legacy function that we need to keep track in the future when we are going to upgrade django.

mozillians/templates/phonebook/emails.html
```diff
@@ -0,0 +1,57 @@
...
+    {{ csrf() }}
+
+    <h2>Primary e-mail address</h2>
+    <fieldset>
+      <span>{{ primary_email }}</span>
+      {{ privacy_field(email_privacy_form.privacy_email) }}
+    </fieldset>
+
+    <h2>Secondary e-mail addresses</h2>
+    <fieldset id="alternate_email">
```
akatsoulas: Probably it's better to use dashes instead of an underscore. This has multiple occurrences.

mozillians/static/mozillians/css/main.less
```diff
@@ -2079,3 +2080,30 @@ body#betasearch {
...
     min-height: 90px;
 }
+
+/* Email settings
+================== */
+#add_email {
+    display: inline-block;
+}
+
+#emails{
```
akatsoulas: Is this used anywhere?

mozillians/phonebook/views.py
johngian: I added this to make sure that both objects are saved in the DB and in case something goes wrong happens we don't make and DB changes.

johngian: Thanks for pointing out that these are legacy functions. We can use the new ones if available in our current version.

mozillians/static/mozillians/css/main.less
johngian: Yes, in the delete buttons. ``#emails`` is defined here:

https://github.com/mozilla/mozillians/pull/1144/files#diff-a141e58a9b2bad6e8b2710227bb4bf7eR4

johngian:
@akatsoulas @comzeradd PR updated

comzeradd:
Looks good. r+

mozillians/templates/phonebook/profile.html
```diff
@@ -112,7 +112,7 @@
               <section class="u-email">
                 <i class="icon-envelope-o"></i>
                 {% if user != shown_user %}<a href="mailto:{{ profile.email }}">{% endif %}
-                  <span class="email">{{ profile.email }}</span>
+                  <span class="email">{{ profile.email|urlize }}</span>
```
glogiotatidis: I think we have logic in 114 and 116 to avoid urlizing all emails. Do I miss something?

johngian: I didn't know that this was intended or something that fell through the cracks. For me, it makes sense to have a ``mailto`` link in when we want users to communicate with a Mozillian. Why did we decided that in the first place? I assume spam emails right?

johngian:
@glogiotatidis PR updated

PR #1149, opened 2015-01-14
Upgrade to latest playdoh-lib

johngian:
* Upgrade Django to 1.6.10
* Upgrade django-compressor to v.1.4
* Remove django-jingo-offline-compressor dependency (not supported by latest django-compressor)
* Use django-compressor for offline compression

### DIFF
--- a/mozillians/phonebook/forms.py
+++ b/mozillians/phonebook/forms.py
@@ -91,7 +91,7 @@ class SearchFilter(django_filters.FilterSet):
     )
 
     vouched = django_filters.ChoiceFilter(
-        name='vouched', label=_lazy('Display only'), required=False,
+        name='vouched', label=_lazy(u'Display only'), required=False,
         choices=CHOICES, action=filter_vouched)
 
     class Meta:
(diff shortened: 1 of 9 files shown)


akatsoulas:
I believe all the changes for the update to Django version 1.6.10 should be in one commit in order to easily revert if needed.

akatsoulas:
Update the name of this PR to something more descriptive like 'Update to latest playdoh-lib and to Django version 1.6.10' or similar

johngian:
I think that's better to handle dependency changes in different commits than the codebase changes.

akatsoulas:
Although it's easier to manage PRs this way, it's not easy to revert if something goes wrong. I would favor multiple commits if there were multiple updates of multiple packages, independent of each other. Regarding the changes in the codebase, the only reason they exist is because they are needed for the update to version 1.6.10. If this not the case, then please move them to another PR. Although I don't agree for the reasons explained, it's up to you whether you are going to proceed with multiple commits or not.

mozillians/phonebook/forms.py
```diff
@@ -157,12 +157,12 @@ class ProfileForm(happyforms.ModelForm):
     lat = forms.FloatField(widget=forms.HiddenInput)
     lng = forms.FloatField(widget=forms.HiddenInput)
     savecountry = forms.BooleanField(
-        label=_lazy('Required'),
+        label=_lazy(u'Required'),
```
akatsoulas: Do you need this for the update?

johngian: Yes

mozillians/phonebook/validators.py
```diff
@@ -71,8 +71,13 @@ def validate_username_not_url(username):
 
 def validate_email(value):
     """Validate that a username is email like."""
-    if not email_re.match(value):
+    _validate_email = EmailValidator()
```
akatsoulas: Why not use the `validate_email` instance?

mozillians/phonebook/forms.py
johngian: Yes, without this ``_lazy`` raises ``TypeError("Lazy object returned unexpected type.")``.

akatsoulas:
I tested this locally and everything seems to be working fine. r+wc

mozillians/phonebook/validators.py
johngian: To keep it consistent with the rest of the validators. We could use validate_email and import with a different name, but it does pretty much the same thing I am doing here.

johngian:
Squashed commits to a single one.

PR #1151, opened 2015-01-16
Use QuerySet.datetimes() instead of QuerySet.dates().

johngian:


johngian:
@akatsoulas r?

akatsoulas:
lgtm. Just make sure before merging that the db in all server instances can support this. r+ :coffee:

PR #1137, opened 2014-12-22
1031286 location wrong

wraybowling:
Fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1031286

(will squash once verified)

PR #1156, opened 2015-02-03
[fix bug 1031286] Let users know how they can fix wrong geolocation data...

comzeradd:
This is a small improvement. [Fixes #1137]


johngian:
Code looks good. I don't really like that we use both parenthesis and brackets. Here is my take on the UI:

https://github.com/johngian/mozillians/commit/003090ef2ae87c3ba04de75f8dfcffe555fbd643

comzeradd:
Thanks for the suggestion. I updated the PR based on your feedback.

PR #1157, opened 2015-02-04
Add django-rest-framework to requirements.

akatsoulas:


johngian:
I am not sure why we need this. We already maintain ``django-rest-framework`` dependency in ``vendor-local``.

akatsoulas:
I would say, with the current setup, just for reference. It's like django-import-export. Although it's installed in `vendor-local/lib`, it's also in the requirements file.  On the other hand, since there is no actual need for this, feel free to close it.

johngian:
Let's merge that since this is going to be helpful if we want to start using peep.

johngian:
r+

PR #1161, opened 2015-02-04
Use consistent border-color for both inputs and textareas

comzeradd:
Currenly textareas have a lighten border color. (See Bio at [edit profile](https://mozillians.org/en-US/user/edit/) for example). This is caused because we overwrite sandstone's default.


johngian:
Please reference this [1] bug without adding the word fix, just add ``[Bug <>]`` prefix.

johngian:
Please open a bug to track this fix.

johngian:
r+wc

comzeradd:
commit amended

PR #1171, opened 2015-02-24
[Fix bug 926673] Use privacy aware attributes instead of queries.

johngian:
* Add WEBSITE external account field.


johngian:
@akatsoulas r?

akatsoulas:
Code looks good. r+

PR #1175, opened 2015-03-03
[Fix bug 1131009] Change wording for API keys management.

johngian:


johngian:
@comzeradd r? please

mozillians/templates/phonebook/edit_profile.html
```diff
@@ -453,68 +453,9 @@ <h2 class="required">{{ _('Your location') }}</h2>
...
-              <li>{{ app.name }} {% if app.url %}<em>({{ app.url|urlize }})</em>{% endif %}<br>
-                <small><strong>{{ _('Key') }}:</strong> {{ app.key }}</small>
-              </li>
-            {% else %}
-              {{ _('You have no registered, active apps.') }}
-            {% endfor %}
-          </ul>
+        <h4>{{ _('Developers') }}</h4>
+        <p>Developers can request API keys and manage existing ones using our dedicated API page.</p>
+        <a class="cancel" href="{{ url('phonebook:apikeys') }}">Manage API keys</a>
```
akatsoulas: Maybe just a plain link, without the `cancel` class is more inline with the rest of the page.

akatsoulas: If I click on the link, I get a 404 error.

johngian: We decided this to be a ``cancel`` button with @ppapadeas.
@ppapadeas what do you think?

mozillians/templates/phonebook/edit_profile.html
```diff
@@ -437,7 +437,7 @@ <h2 class="required">{{ _('Your location') }}</h2>
           and sites like <a id="services-bugzilla-url" href="{{ bugzilla_url }}">Bugzilla</a> and the
           <a id="services-mozilla-reps" href="{{ reps_url }}">Mozilla Reps Portal</a>.
         {% endtrans %}
-      <h4>{{ _('API Version 1') }}</h4>
+      <h4>{{ _('Privacy') }}</h4>
```
akatsoulas: Users might find it difficult to understand what these privacy fields are about, without some reference to the API v1.

mozillians/templates/phonebook/edit_profile.html
johngian: For the 404 you need to open a waffle flag. Also this is temporary until we have proper documentation for the API.

akatsoulas: Then let's make the link/section visible only if the waffle flag is active.

ppapadeas: On button, it is nicer as a button. :)
On the permission denied it is fine, we just need to change the wording of the link to say "browsable API" and not Documentation. Once the documentation is up then we fix the wording and the link. @johngian r+wc

mozillians/templates/phonebook/edit_profile.html
ppapadeas: It is mentioned above. (usage of API by other sites)

akatsoulas:
r+wc :beer:

mozillians/templates/phonebook/edit_profile.html
akatsoulas: It doesn't specify what it's specifically for (v1 or v2) - it just displays two privacy fields regarding api. Personally, I find this confusing and I would suggest a more meaningful heading but it's up to you.

PR #1177, opened 2015-03-07
Add contribute.json file.

Osmose:
Add a contribute.json file a la http://contribute.paas.allizom.org/.

### DIFF
--- a/contribute.json
+++ b/contribute.json
@@ -1,0 +1,30 @@
+{
+    "name": "Mozillians",
+    "description": "A community directory for Mozillians to connect with each other.",
+    "repository": {
+        "url": "https://github.com/mozilla/mozillians",
+        "license": "BSD 3-Clause",
+        "tests": "https://travis-ci.org/mozilla/mozillians"
+    },
+    "participate": {
+        "home": "http://mozillians.readthedocs.org/en/latest/contribute.html",
+        "docs": "http://mozillians.readthedocs.org/en/latest/installation/installation.html",
+        "irc": "irc://irc.mozilla.org/#mozillians",
+        "irc-contacts": ["nemo-yiannis"]
+    },
+    "bugs": {
+        "list": "https://bugzilla.mozilla.org/buglist.cgi?product=Participation%20Infrastructure;component=Phonebook;resolution=---",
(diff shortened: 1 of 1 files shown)


Osmose:
@johngian r?

johngian:
Hey @Osmose 
We already provide a ``contribute.json`` here [1][2]. It might need to get a small update though.
Thanks for your PR :+1: 

[1] https://mozillians.org/contribute.json
[2] https://github.com/mozilla/mozillians/blob/91ed5e3eeb1f926d7864ea75a3a8a4f562a659f7/mozillians/humans/views.py#L15

Osmose:
So you seem to be in the same position as [fjord](https://github.com/mozilla/fjord/pull/508#issuecomment-77074725): while putting contribute.json at the root of the website is a useful step, http://contribute.paas.allizom.org/ says they should be put at the root of a repo to be consistent with libraries and other non-website projects that have no domain to put them under.

I've been idly playing around with the idea of making a library to support putting the contribute.json file at the root of the repo and serving it up as well; would you be interesting in using a library like that?

Osmose:
(This question applies to ReMo as well, as I noticed it's in the same position.)

johngian:
@Osmose I don't have any strong feelings to serve ``contribute.json`` from the root of the repo. We can move it there if this is required from other libraries.

Quick question: Do we need to both serve ``/contribute.json`` from Mozillians.org and have it at the top of the repo?

Osmose:
> Quick question: Do we need to both serve /contribute.json from Mozillians.org and have it at the top of the repo?

The only stated requirement on the contribute.json file is that it be at the root of the repo. Serving it up at `/contribute.json` was something a few sites just did of their own accord, but now some tools (mainly https://github.com/mhoye/muditanist) are assuming that websites will serve it up.

IMO doing both is good, but only having it at the root of your repo is _required_.

johngian:
Closing this one after #1182 where:
* We are moving ``contribute.json`` to the root of the repo
* Update some entries
* Fix view to serve the ``contribute.json`` from the new path.

Thank you @Osmose :rocket:

Osmose:
<3

PR #1178, opened 2015-03-11
[Fix bug 1130389] Fix UNIQUE constraint violation for "City".

johngian:


johngian:
@akatsoulas r?

akatsoulas:
I am not sure that the de-duplication process should take place in each lookup request. Thoughts?

johngian:
We could spawn a task instead, but it's not that much of an overhead doing 2 HTTP requests to mapbox. We could implement this using the following steps:

* Check for duplicates based on ``lookup_args``.
* If duplicate is found, use this one and spawn a deduplication task to run in the background.

johngian:
FYI this has happened for 2-3 entries until now, so we are talking about an overhead of ~10 requests in total. If the duplicates are fixed we will be running the same code we dow ATM.

mozillians/geo/lookup.py
```diff
@@ -177,10 +225,25 @@ def result_to_city(result, country, region):
...
+                duplicate_city_keys = get_city_keys(duplicate_city.id)
+
+                if all([city_keys, duplicate_city_keys, duplicate_city_keys == city_keys]):
+                    deduplicate_cities([city.id, duplicate_city.id])
+
         except City.DoesNotExist:
             # Mapbox sometimes returns multiple cities with the same
             # name but different ids. So we need to check for existing
-            # (name, region, country) groups to avoid filling the
+            # (name, region, country) groups to avoid filling thge
```
akatsoulas: Small typo here

mozillians/geo/lookup.py
```diff
@@ -177,10 +225,25 @@ def result_to_city(result, country, region):
 
         try:
             city = City.objects.get(mapbox_id=mapbox_city['id'])
+
+            # Check for duplicates
+            qs = City.objects.filter(**lookup_args).exclude(mapbox_id=city.mapbox_id)
```
akatsoulas: You can simplify this with something like:

```python
try:
    city = City.objects.get(**lookup_args).exclude(mapbox_id=city.mapbox_id)
except City.DoesNotExist:
    pass
```

then you can pass the object to your helper functions directly, avoiding a couple of queries

johngian:
@akatsoulas PR updated after your suggestions

mozillians/geo/lookup.py
```diff
@@ -194,5 +207,16 @@ def result_to_city(result, country, region):
...
                     do_save = True
             if do_save:
-                city.save()
+                try:
+                    city.save()
+                except IntegrityError as e:
+                    # Check if duplicate City entry with the same
+                    # (name, region, country) exists
+                    qs = City.objects.filter(**lookup_args)
+                    if qs.exists():
```
akatsoulas: I would suggest to move the following code outside of the `except` statement.

mozillians/geo/lookup.py
```diff
@@ -194,5 +207,16 @@ def result_to_city(result, country, region):
...
+                        # Single duplicate exists because of unique_together
+                        deduplicate_cities(city, qs[0])
```
akatsoulas: Since this function only performs a couple of queries I would suggest to add these operations here and remove completely the helper function to improve the readability of the code. I don't have strong feelings about this though.

johngian: It's easier to test it this way. I suggest to live it as it is.

mozillians/geo/lookup.py
```diff
@@ -177,12 +190,12 @@ def result_to_city(result, country, region):
 
         try:
```
akatsoulas: On another notice the whole block of code could be simplified to something like:

```python
if City.objects.filter(**lookup_args).exclude(mapbox_id=mapbox_city['id']).exists():
    # we have duplicate
    deduplicate(args)
else:
     city, created = City.objects.get_or_create(defaults=defaults, **lookup_args)
     if not created:
          city.name  = lookup_args['name']
          city.save()
return city
````

Of course this needs some fine tuning. It's totally up to you if you wish to follow this approach.

**Comment by akatsoulas** [comment]
r+wc :beers:

File: mozillians/geo/lookup.py [hunk 6]
```diff
@@ -155,6 +156,18 @@ def result_to_region(result, country):
         return region
 
 
+def deduplicate_cities(city, dup):
```
akatsoulas: s/dup/dup_city/g or something similar

mozillians/geo/lookup.py
```diff
@@ -155,6 +156,17 @@ def result_to_region(result, country):
         return region
 
 
+def deduplicate_cities(city, dup_city):
+    """
+    Given 2 City instances, deduplicate dup_city and it's references to city.
+    """
+    users = dup_city.userprofile_set.all()
+    users.update(geo_city=city, geo_region=city.region, geo_country=city.country)
```
akatsoulas: You can reduce the queries by rewriting this to something like:

```python
if not users:
    dup_city.delete()
else: 
    users.update(geo_city=city, geo_region=city.region, geo_country=city.country)

akatsoulas:
I tested this locally and the problem with the unique constraint is fixed when there is the same mapbox id in the lookup and in our db for different regions.
Really nice work! r+wc :rocket:

PR #1180, opened 2015-03-18
Add vouched status in detailed userprofile API serializer.

johngian:


akatsoulas:
Since this is for the API, I would suggest to open a bug so that we can track and document every change on it.

johngian:
Bug opened here [1]
@akatsoulas r?

https://bugzilla.mozilla.org/show_bug.cgi?id=1145696

akatsoulas:
r+

PR #1182, opened 2015-03-19
Serve contribute.json from the root of the repo.

johngian:


contribute.json
```diff
@@ -11,10 +11,10 @@
         "mailing-list": "https://lists.mozilla.org/listinfo/dev-community-tools",
         "irc": "irc://irc.mozilla.org/#commtools",
         "irc-contacts": [
-            "giorgos",
             "nemo-yiannis",
             "tasos",
-            "williamr"
+            "pierros",
+            "comzeradd"
```
comzeradd: I've changed my irc nick to `nikos` :)

johngian:
@akatsoulas r?

mozillians/humans/views.py
```diff
@@ -14,5 +14,4 @@ def humans(request):
 @allow_public
 def contribute_view(request):
     """Generate a contribute.json"""
-    template = render(request, 'humans/contribute.json')
-    return HttpResponse(template, mimetype='application/json')
+    return serve(request, 'contribute.json', document_root=settings.ROOT)
```
akatsoulas: You can serve this directly from the appropriate urls.py and remove the view.

akatsoulas:
r+wc :beetle:

mozillians/humans/views.py
johngian: In our case we need to explicitly ``@allow_public`` to allow unauthenticated users to view ``contribute.json``.

glogiotatidis: tip: you can still do this without a view https://github.com/mozilla/mozillians/blob/master/mozillians/phonebook/urls.py#L48-L53

mozillians/humans/views.py
```diff
@@ -14,5 +14,4 @@ def humans(request):
 @allow_public
 def contribute_view(request):
     """Generate a contribute.json"""
```
glogiotatidis: nit: we don't generate, we just serve

PR #1186, opened 2015-03-31
[Fix bug 1149451] Add celery check button in /admin.

johngian:


johngian:
@akatsoulas r?
It's pretty the same functionality we have in ``remo``.

akatsoulas:
r+

PR #1188, opened 2015-04-01
Docker development environment maintenance

johngian:


comzeradd:
I'm concerned that maybe it's too early to switch to docker-compose, since it's not yet packaged for any linux distribution.

johngian:
We are heavily dependent on pypi, so for me it doesn't make any difference to install it via ``pip`` instead of using the system's package manager.

comzeradd:
I think that we are not using `pip` at all for the [Docker installation](http://mozillians.readthedocs.org/en/latest/installation/installation-docker.html).

johngian:
Yes, but we are redirecting to fig's installation that instructs to either manually ``curl`` script to path (which is bad) or install via pip. I mentioned that we depend on pip because part of our installation environment is based on our ``requirements/*.txt``.

comzeradd:
Looks good. r+

PR #1191, opened 2015-04-14
Add missing API v2 filters

johngian:
* Language
* Account
* Group
* Skill

### DIFF
--- a/mozillians/users/tests/test_api_v2.py
+++ b/mozillians/users/tests/test_api_v2.py
@@ -7,8 +7,9 @@ from nose.tools import eq_, ok_
 
 from mozillians.common.tests import TestCase
 from mozillians.geo.tests import CityFactory, CountryFactory, RegionFactory
+from mozillians.groups.tests import GroupFactory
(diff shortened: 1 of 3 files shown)


johngian:
@akatsoulas can you take a look at this?

akatsoulas:
@johngian sure!

mozillians/users/tests/test_api_v2.py
```diff
@@ -196,3 +197,22 @@ def test_filter_emails_alternate(self):
         f = UserProfileFilter(request.GET, queryset=UserProfile.objects.all())
         eq_(f.qs.count(), 1)
         eq_(f.qs[0], user.userprofile)
```
akatsoulas: Shouldn't you add a few more tests for the rest of the filters?

johngian: We are not doing something else custom except of this so I suppose we don't need to check ``django-filters`` functionality nor ``django lookups``.

akatsoulas:
Looking good r+

PR #1195, opened 2015-05-27
[Fix bug 1168765] Include API associated emails in /admin list and ex…

johngian:
…port

### DIFF
--- a/mozillians/api/admin.py
+++ b/mozillians/api/admin.py
@@ -1,26 +1,62 @@
 from django.contrib import admin
 
 import autocomplete_light
+from import_export import fields
 from import_export.admin import ExportMixin
+from import_export.resources import ModelResource
 
 from models import APIApp, APIv2App
 
 
+class APIAppResource(ModelResource):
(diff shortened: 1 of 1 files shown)


mozillians/api/admin.py
```diff
@@ -1,26 +1,62 @@
...
+
+
 class APIAppAdmin(ExportMixin, admin.ModelAdmin):
     """APIApp Admin."""
-    list_display = ['name', 'key', 'owner', 'is_mozilla_app', 'is_active']
+
+    list_display = ['name', 'key', 'owner', 'owner_email', 'is_mozilla_app', 'is_active']
     list_filter = ['is_mozilla_app', 'is_active']
+
+    def owner_email(self, obj):
```
akatsoulas: Since you added the email in the APIAppResource, why do you need this?

akatsoulas: Personally, I don't think that you need to display the email in the list_display, but it's totally up to you.

akatsoulas:
r+

PR #1197, opened 2015-05-27
[fix bug 1167770] Fix css for adding external accounts in mobile.

akatsoulas:


comzeradd:
It work great local. r+

PR #1200, opened 2015-06-28
[fix bug 1133122] Move the search bar to the top.

brian-yang:


akatsoulas:
@comzeradd maybe you can do a review on this?

comzeradd:
So this change is for the mobile view only. But the proposed change breaks the desktop view (see below). I would suggest you create a new id just for search (don't duplicate `content-info`) and use media queries to adjust it for desktop/mobile view. For instance on desktop you probably want to use `float: right` to make it move to the proper position.

home

PR #1201, opened 2015-06-29
[fix bug 1178348] - wrong link to Mozilla Locamotion in mozillians.org profiles

pollti:


akatsoulas:
r+ thanks for fixing this!

PR #1204, opened 2015-07-09
[Fix bug 1170046] Update example mapbox API token.

johngian:
This fixes issues in local/dev/stage mozillians.org environments.


akatsoulas:
r+

PR #1206, opened 2015-07-16
Upgrade django to version 1.7.10

johngian:


akatsoulas:
I would remove both south and migrations in one commit since both commits need each other in order to work in case of reverting back

akatsoulas:
I don't know if you are interested in something like this, but this seems like a good opportunity to remove playdoh altogether along with any monkey patches there are in the codebase.

akatsoulas:
If you don't remove playdoh, I don't think that there are much to review here. If you feel confident with this you can proceed and merge it. I would organize the commits differently (into logical blocks) but this is up to you. My only question is about data migrations, do they need any special treatment or you just remove them completely?

johngian:
@akatsoulas Thanks for the reviews. I was also thinking to squash some of the commits in order for them to be more atomic. I opened the PR like this in order to show all the steps in order to migrate to 1.7.

johngian:
I suggest we merge this without removing playdoh. We can tackle legacy playdoh code in another PR.

johngian:
Some things that came up while testing this PR:

* We need to make ``AppConfig``s for each mozillians app (already fixed it locally)
* Admin is not working ATM in this PR (need to fix it)

johngian:
@akatsoulas PR updated with some bug squashes and fixes to the previous comment.

PR #1207, opened 2015-07-21
Add missing migration in mozspaces app.

johngian:


johngian:
@akatsoulas This is required in order to have our migrations in sync before moving to django's native migrations in 1.7.

Can you please verify that this is missing?

akatsoulas:
I run the migrations on a clean db and it seems that there's a missing migration in mozspaces app. Nice catch r+ :+1:

PR #1212, opened 2015-08-04
[Fix bug 1183220] Fix policy for public indexable profile.

johngian:
* Add to public index only when all PUBLIC_INDEXABLE_FIELDS are public.getattr(self, field, None)


johngian:
@akatsoulas r?

glogiotatidis:
I'm not sure that this is what you want. Back then we decided that we need at least one of the `PUBLIC_INDEXABLE_FIELDS` to be public in order to add the profile to the public index. The spirit is that `PUBLIC_INDEXABLE_FIELDS` are fields that can identify the person (i.e. name, email). If the person hasn't made any of these fields public it makes no sense to make the profile searchable. It would be like I make my `bio` public but not anything else.

Note that this is about indexing, profiles with any public field are made public exposing only the public fields.

glogiotatidis:
also can you give me access to bug 1183220 @johngian ?

johngian:
@glogiotatidis I cc'ed you in the bug

johngian:
Thanks for noticing @glogiotatidis . Indeed this was a regression introduced in this [1] commit. Closing this PR and will open a new one with the proper fix.

:+1: 

[1] https://github.com/mozilla/mozillians/commit/593ee6eb885e692cdc4bf2d94c25bae668c361f9#diff-5c2c0011d4d204047aa26eedfff367bbR28

PR #1216, opened 2015-08-05
Add an editorconfig file.

akatsoulas:


.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
+
+# Python configuration
+[*.py]
+charset = utf-8
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
```
comzeradd: You can have a `[*]` section for common things. (trailing white space, charset, space for indent). You can also add an `end_of_line = lf` just ot be safe.

.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
+
+# Python configuration
+[*.py]
```
comzeradd: You can also add a css/less here, where we also use 4 spaces.

akatsoulas:
@comzeradd pr updated. Thanks for the review!

comzeradd:
r+

PR #1220, opened 2015-08-10
Profile UI redesign

comzeradd:


johngian:
Some notes here after [1]
* Full name seems to be misaligned with the rest of the fields.

selection_030

* Favicon in ``settings`` page is not working (seems like a regression, master works fine for me)
* Skills beautification seems broken as well

selection_031

* Email section needs some love
 * Issue in persona button rendering
 * Too much space left between primary and secondary emails
 * Mixed bold and plain titles
selection_032

 Except of these there are also some leftovers from [1] eg.
* Location permissions UI
* Contribution story

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1129848#c3

johngian:
Also can you split the top commit to more atomic/smaller ones. There are massive changes and its not easily reviewed and maintained.

johngian:
I am getting this error on new profile creation:
``'location_privacy_field' is undefined``

I cannot find a reference in your branch or master. Is that something you forgot to commit?

johngian:
We also need to include this field in new user registration
selection_033

johngian:
* Default delete email button looks like this.

selection_034

johngian:
* We also need to fix the alignment in case we don't refactor the permission level UI.

selection_035

johngian:
After form submission (eg. in case of an error) privacy toggles are not getting the value that the hidden (real) input element has.

johngian:
* On mobile view location search field overflows map.

selection_037

johngian:
If you haven't fixed these yet please cherry-pick this [1]
These are minor fixes to make the branch work when I started working for the backend.

[1] https://github.com/johngian/mozillians/commit/c382cfd7d9b409270dda8ba88d8f4dfa86f3970c

akatsoulas:
I think `Skills` section should be moved away from the `profile` and go together with `Groups` and rename the menu to something like `Groups & Skills`. It's a better logical fit there and skills are not basic information.

johngian:
Languages can also go there, since that's a skill as well.

akatsoulas:
The `Profile` section should have only one update button. All the fields there should be posted into one form and not in multiple ones

akatsoulas:
The same is true for all the sections. Since we are breaking the profile into sections there is no need to post multiple times per section

mozillians/templates/phonebook/includes/profile_edit_basic.html
```diff
@@ -0,0 +1,243 @@
...
+          <input type="checkbox" class="acknowledge">
+          {% trans %}
+            I acknowledge that deleting my account will make it tougher
+            for me to show off my contributions and easily be found by
+            other members of the Mozilla community.
+          {% endtrans %}
+        </fieldset>
+      </div>
+      <div class="panel-footer">
+        <a href="{{ url('phonebook:profile_confirm_delete') }}" class="btn btn-danger delete">
```
akatsoulas: This link is clickable even if the user does not select the checkbox above.

mozillians/templates/phonebook/includes/profile_edit_basic.html
```diff
@@ -0,0 +1,243 @@
...
+        <button class="btn btn-primary" id="form-submit-top" type="submit">{{ _('Update Languages') }}</button>
+      </div>
+    </div>
+  </form>
+
+  <form class="edit-profile" method="POST" action="{{ url('phonebook:profile_edit') }}?next=#profile" enctype="multipart/form-data">
+    {{ csrf() }}
+    <div class="panel panel-danger">
+      <div class="panel-heading">
+        {{ _('Delete account') }}
```
akatsoulas: I would suggest not to change the strings if there isn't a need to because new strings need to be re-localized.

comzeradd:
PR updated

Things fixed:
* Break down to more commits
* Fix favicon bug due to `location.hash`
* Fix spacing, bold mixup on email settings
* Wrong reference to non-existant privacy field
* Add Privacy Opt-in for new users
* Fix delete buttons opacity
* Fix overflow on mobile location settings
* Fix alignment on location fields

Things not fixed yet:
* Group privacy settings on Location
* Add placeholder to contribution story
Both of these can come later on a separate commit

Things that need some more work:
* Mobile view needs some love with spacing and alignments
* Make sure the 'Delete Profile' button is not clickable untill user ticks the checkbox
* Privacy button not taking stored value on form submission (I need the backend to re-produce this)

johngian:
* Misaligned fields still exist (``username`` is next to label, ``full name`` is below)
selection_039

johngian:
* Functional areas in mobile view is broken
selection_040

johngian:
* The same happens for groups
selection_041

johngian:
FYI I pushed here the rebased branch which combines both frontend and backend in case you needed it:
https://github.com/johngian/mozillians/commits/1191850-with-ui

johngian:
This PR also breaks the ``groups`` UI:
selection_042

johngian:
* City/region/country save button and name are misaligned in mobile view.

selection_043

johngian:
In mobile view we are providing a button to show all settings but while browsing I would expect to also have a button to get you back to the ``restricted`

johngian:
We used to have a template to include for the photo form [1]. Please update it with the new UI and use it in both registration/basic information forms.

[1] mozillians/templates/phonebook/includes/photo_form.html

johngian:
This [1] template looks like its not used. Please double check and remove it if orphaned.

mozillians/templates/phonebook/includes/photo.html

johngian:
When I open a link directly in mobile view (not #profile) like:
http://127.0.0.1:8000/en-US/user/edit/#location

It shows this first:
selection_044

And when I click "show all" I get an empty page:
selection_045

comzeradd:
> We used to have a template to include for the photo form [1]. Please update it with the new UI and use it in both registration/basic information forms.
> 
> [1] mozillians/templates/phonebook/includes/photo_form.html
>
> This [1] template looks like its not used. Please double check and remove it if orphaned.
> 
> [1] mozillians/templates/phonebook/includes/photo.html

It seems that these are not used on the current code either. Can you point me to the code that includes these two templates?

johngian:
We are not including them at the moment but we should since photo form is rendered twice (both in registration and basic information form). In the previous comment I suggest to move the ``photo_form`` functionality to a different template and include it twice.

About ``photo.html`` we are not using so we can remove it.

mozillians/static/mozillians/js/profile_edit.js
```diff
@@ -1,4 +1,57 @@
 $(function() {
+    // Favicon
```
johngian: Please ``"use strict";`` here.

mozillians/static/mozillians/js/profile_edit.js
```diff
@@ -1,4 +1,57 @@
...
+    $('.settings-nav > li > a').on('click', function(e) {
+        e.preventDefault();
+        $('.settings-tab').hide();
+        $('.settings-nav > li').removeClass('active');
+        $(this).parent('li').addClass('active');
+        var content = $(this).attr('href');
+        $(content + '-tab').show();
+        location.hash = content;
+    });
+    if (screen.width < 992) {
```
johngian: Please move ``992`` to a constant with a comment regarding what it does (I assume bootstrap device size detection).

mozillians/static/mozillians/js/profile_edit.js
```diff
@@ -1,4 +1,57 @@
...
+        $(this).hide();
+        $('.settings-tab').removeClass('hidden-sm');
+        $('.settings-tab').removeClass('hidden-xs');
+    });
+
+    // Privacy toggle buttons
+    $('.privacy-toggle').on('click', function(e) {
+          e.preventDefault();
+          var firstchoice = $(this);
+          var secondchoice = $(this).closest('.btn-group').find('.privacy-toggle');
```
johngian: How do you ensure that ``$(this).closest('.btn-group').find('.privacy-toggle')`` this doesn't return ``$(this)``? Does jquery exclude this?

mozillians/static/mozillians/js/profile_edit.js
```diff
@@ -1,4 +1,57 @@
...
+          secondchoice.removeClass('active');
+          firstchoice.toggleClass('active');
+          var target = $(this).closest('.btn-group').find('input');
+      if (firstchoice.hasClass('active')) {
```
johngian: You could avoid this ``if/else`` by getting directly the active element's val.

johngian: Else you could do something like this:
``
choice = (firstchoice.hasClass('active') ? firstchoice : secondcoice)
target.val(choice.data('value'))
``
The idea is that code in if/else branches is way similar to repeat.

mozillians/static/mozillians/js/profile_edit.js
akatsoulas: Is this needed at all?

comzeradd:
@johngian PR updated

* Fix Groups and Functional Areas grid
* Fix Profile misalignment on Full Name
* Fix 'Delete Profile' button to be inactive until checkbox is checked
* I moved profile photo to another template
* I made some code changes on privacy toggles based on your comments

akatsoulas:
I am not sure if this is commented somewhere, but when a new user registers a new profile, the privacy button needs to be set to a default value (I would go with `Mozillians`).

mozillians/static/mozillians/js/profile_edit.js
```diff
@@ -1,4 +1,58 @@
...
+    $('.settings-nav > li > a').on('click', function(e) {
+        e.preventDefault();
+        $('.settings-tab').hide();
+        $('.settings-nav > li').removeClass('active');
+        $(this).parent('li').addClass('active');
+        var content = $(this).attr('href');
+        $(content + '-tab').show();
+        location.hash = content;
+    });
+    if (screen.width < mobile_breakdown) {
```
akatsoulas: Why is this needed here? Isn't bootstrap handling this?

comzeradd: It's needed for the cases where a user opens a url with a `location.hash` (eg. #groups) that it's not visible on mobile.

johngian:
Closing this one after #1243

PR #1222, opened 2015-08-13
[Fix bug 1193967] Implement pronoun field

TannerFilip:
I'm not positive if I did this right, but it seems to be working locally.


johngian:
Hey @TannerFilip 
There is an ongoing profile refactoring as you can see in PRs #1220 #1242.
I suggest we close this for now and revisit it after we merge the first iteration on the new ``settings`` UI. The reason I am suggesting this is that both forms and templates are changing radically and this PR will require many changes in order to be compatible.

Thank you for your contribution anyway!
Keep on rocking :metal: 

:rocket:

PR #1225, opened 2015-08-19
Preparatory step for Django 1.7x upgrade

johngian:


akatsoulas:
What about using just the `python` executable?

akatsoulas:
I would go with one commit, feel free to ignore though.

akatsoulas:
If you feel confident with this, ship it! r+ :beer:

johngian:
I am not sure which is the default python binary in our cluster and I prefer it to be explicit. Also other projects using chief do it the same way [1] .

[1] https://github.com/mozilla/kitsune/blob/master/scripts/update/deploy.py

PR #1229, opened 2015-08-21
Update wsgi script

akatsoulas:


johngian:
We don't need to set this env var for the chief script runtime.

johngian:
I suggest we close this for now.

PR #1231, opened 2015-08-24
Fix slow queries

akatsoulas:


johngian:
WOOOOT!

johngian:
We also need to make sure all managers are up to date with 1.7 convention:
```
mozillians/groups/managers.py
mozillians/announcements/managers.py
mozillians/users/managers.py
```

johngian:
Please include the bug number to the commit message.

mozillians/users/models.py
```diff
@@ -135,7 +135,7 @@ class UserProfile(UserProfilePrivacyModel):
         ('contribute', 'Get Involved'),
     )
 
-    objects = UserProfileManager()
+    objects = UserProfileQuerySet.as_manager()
```
johngian: You might also need this here to be compatible with the previous manager.
```
     objects = UserProfileQuerySet.as_manager()
+    objects.use_for_related_fields = True
```

johngian:
Please remove the (now) unused class definition of ``UserProfileManager``.

johngian:
The same applies for ``GroupManager``.

akatsoulas:
@johngian pr updated

akatsoulas:
@glogiotatidis since you were the initial author of this, maybe you can have a look too?

johngian:
Let's change ``announcements/{models,managers}.py`` to use the same convention.

akatsoulas:
We are using the default queryset in `announcements` so probably it's OK the way it is now.

johngian:
Code looks good to me (we have already discussed most of these before the PR) :+1:

johngian:
r+ from me :rocket: 

@glogiotatidis 
Any comments?

akatsoulas:
The need for these changes first introduced when we realized that the dev server is unresponsive. It turns out that it was due to really slow queries (~2000ms vs ~70ms before the merge). After debugging the problem lies in the custom manager in UserProfile.

Some useful documentation can be found here:
https://docs.djangoproject.com/en/1.8/topics/db/managers/#creating-manager-with-queryset-methods

johngian:
Please change the commit message to something more descriptive (eg. mention django 1.7 release notes, and the performance issue).

mozillians/users/models.py
```diff
@@ -135,7 +135,7 @@ class UserProfile(UserProfilePrivacyModel):
         ('contribute', 'Get Involved'),
     )
 
-    objects = UserProfileManager()
+    objects = UserProfileManager.from_queryset(UserProfileQuerySet)()
```
glogiotatidis: Have you tried
`objects = UserProfileQuerySet.as_manager()`

and completely remove `UserProfileManager` from `managers.py`?

mozillians/groups/managers.py
```diff
@@ -6,14 +6,11 @@ class GroupBaseManager(Manager):
     use_for_related_fields = True
     queryset_class = QuerySet
```
glogiotatidis: this is not needed.

mozillians/users/models.py
akatsoulas: Yes, and the results were exactly the same with this PR. The reason we favored this approach is in order to add the `use_for_related_fields` option in a sane way.

https://docs.djangoproject.com/en/1.8/topics/db/managers/#controlling-automatic-manager-types

mozillians/groups/managers.py
```diff
@@ -6,14 +6,11 @@ class GroupBaseManager(Manager):
...
 
-    def get_query_set(self):
+    def get_queryset(self):
         qs = self.queryset_class(self.model, using=self._db)
         qs = qs.annotate(member_count=Count('members'))
         return qs
 
-    def __getattr__(self, name):
-        return getattr(self.get_query_set(), name)
-
```
glogiotatidis: `GroupManager = GroupBaseManager.from_queryset(GroupQuerySet)`

and delete `GroupManager` class

akatsoulas: :+1:

mozillians/groups/managers.py
```diff
@@ -6,14 +6,11 @@ class GroupBaseManager(Manager):
     use_for_related_fields = True
     queryset_class = QuerySet
 
-    def get_query_set(self):
+    def get_queryset(self):
         qs = self.queryset_class(self.model, using=self._db)
```
glogiotatidis: Maybe change this to `qs = super(GroupBaseManager, self).get_queryset()` because django's `get_queryset` is different from what we do here  https://github.com/django/django/blob/master/django/db/models/manager.py#L209

mozillians/users/models.py
glogiotatidis: ok in this case remove https://github.com/mozilla/mozillians/pull/1231/files#diff-5cd44f45451d60a29f0196b7a80a91f5R125 ?

akatsoulas:
@glogiotatidis @johngian pr updated

mozillians/users/managers.py
```diff
@@ -122,8 +122,5 @@ class UserProfileManager(Manager):
 
     use_for_related_fields = True
 
-    def get_query_set(self):
-        return UserProfileQuerySet(self.model)
-
-    def __getattr__(self, name):
-        return getattr(self.get_query_set(), name)
+    def get_queryset(self):
```
johngian: Can't this be omitted since it's returning super directly?

glogiotatidis: not needed?

glogiotatidis:
lgtm r+

PR #1233, opened 2015-08-24
[Bug 1196244] Remove duplicate legacy code.

johngian:
We used duplicate code in groups/{forms,models}.py because of this:
https://code.djangoproject.com/ticket/16986

Now with Django 1.7 model can clean individual fields and there is
no need to clean both `ModelForm` and `Model`.

### DIFF
--- a/mozillians/groups/models.py
+++ b/mozillians/groups/models.py
@@ -26,14 +26,8 @@ class GroupBase(models.Model):
         ordering = ['name']
 
     def clean(self):
-        """Verify that name is unique in ALIAS_MODEL.
+        """Verify that name is unique in ALIAS_MODEL."""
 
-        We have to duplicate code here and in
-        forms.GroupForm.clean_name due to bug
-        https://code.djangoproject.com/ticket/16986. To update when we
-        upgrade to Django 1.7.
-
-        """
         super(GroupBase, self).clean()
         query = self.ALIAS_MODEL.objects.filter(name=self.name)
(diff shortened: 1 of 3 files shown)


johngian:
@akatsoulas r?

mozillians/groups/models.py
```diff
@@ -26,20 +26,15 @@ class Meta:
         ordering = ['name']
 
     def clean(self):
```
akatsoulas: Why don't you let django handle this with the `unique

mozillians/groups/models.py
```diff
@@ -26,20 +26,15 @@ class Meta:
...
-        upgrade to Django 1.7.
-
-        """
         super(GroupBase, self).clean()
         query = self.ALIAS_MODEL.objects.filter(name=self.name)
         if self.pk:
             query = query.exclude(alias=self)
         if query.exists():
-            raise ValidationError({'name': _('This name already exists.')})
+            msg = _('Group with this Name already exists.')
```
akatsoulas: Please use the previous msg

akatsoulas:
Code looks good. If you don't need the `Model.clean()` method you can remove it altogether. If this is needed then just update the msg. r+wc :beers:

mozillians/groups/models.py
johngian: We need this in order to have sane messaging when you add an alias with an existing name.

PR #1237, opened 2015-08-25
[bug 1196244] Remove admin.autodiscover from url conf.

akatsoulas:
* https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#discovery-of-admin-files


johngian:
Tested it locally and admin entries still exists. LGTM
r+ :rabbit:

PR #1239, opened 2015-08-27
Remove legacy code from old vouching system

johngian:
* Set Vouch.date not NULL
* Remove management command that deletes legacy vouches


johngian:
@akatsoulas r?

mozillians/users/models.py
```diff
@@ -674,7 +674,7 @@ class Vouch(models.Model):
     autovouch = models.BooleanField(default=False)
 
     # The back-end can set date null, for migration purposes, but forms cannot.
```
akatsoulas: Please remove this comment

mozillians/users/models.py
```diff
@@ -674,7 +674,7 @@ class Vouch(models.Model):
...
-    date = models.DateTimeField(null=True, default=None)
+    date = models.DateTimeField()
```
akatsoulas: I would suggest to leave this as it is. Since we are ensuring that a date is always set through the forms, we should allow a none value through the admin and migrations.

akatsoulas:
r+wc :bear:

mozillians/users/models.py
johngian: Check comment [1]
This is leftover and dependency in order to close the legacy vouching bug.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1033305#c2

akatsoulas: Then please remove the comment a few lines above :)

PR #1244, opened 2015-09-11
[Bug 1202601] [WIP] Remove playdoh from project dependencies

safwanrahman:
Here is the patch to remove the Playdoh and implement peep for installing python dependencies.

Install all the default python dependencies with ``./peep.sh install -r requirements/default.txt``
and all the dev dependencies with ``./peep.sh install -r requirements/default-dev.txt``

Still some works need to be done!
- [x] Remove the vendor and vendor-local directory
- [x] Remove ``funfactory`` from `./manage.py`
- [ ] Make appropriate changes in deploy script for deploying
- [ ] Update the documentation
- [ ] Configure docker
- [ ] Maybe changes in the server for the virtualenv

@johngian thoughts about this? :smile:


johngian:
Hey @safwanrahman 
Thanks for your contribution. As you already mention in your PR it's still WIP.
Some comments in current changes:

* I don't think that we need to maintain ``dictconfig.py`` since we are not using it
* It doesn't sound like a good idea to maintain ``celeryutils.py`` just to use ``chunked``.
 * We could move this helper to a ``common/utils.py`` file.
* ``./manage.py`` still uses ``funfactory`` and monkeypatches it to import ``vendor``
 * You need to remove these dependencies from manage.py
* At the moment even if ``travis`` build passes you haven't removed the real dependencies' code under ``vendor`` and ``vendor-local``. This might be misleading, since many import will brake.

safwanrahman:
@johngian Review the progress?
Have fixed most of the issue you mentioned in the last comment.
regarding dictconfig, we are using dictconfig here https://github.com/mozilla/mozillians/blob/master/log_settings.py#L68

johngian:
Regarding ``dictconfig`` instead of using this legacy ``dictconfig`` implementation I suggest we adopt the one [1] from django. Here is a similar usage example [2].

[1] https://github.com/django/django/blob/1.7.10/django/utils/dictconfig.py
[2] https://github.com/mozilla/kitsune/blob/adde3c81e34b85a62ff10776686cbb5589a2b180/kitsune/log_settings.py#L4

johngian:
This PR is targeting ``mozilla/mozillians/master" from your master. The usual workflow that we suggest is to work on a branch ``<#Bug number>`` and then open a PR to master.

Since you have already done a lot of work on that functionality I suggest you get assigned to that bug and open a new PR as described.

johngian:
Regarding the commits in this PR:
* Let's squash everything that removes ``vendor`` and ``vendor-local``
* On top of that let's add all the copied code from vendor and the rest of the dependencies
* Finally add a commit with all the ``mozillians`` codebase changes.

It would be much easier to review.

PR #1245, opened 2015-09-17
Fix login_failure regression

johngian:
* Introduced in django-browserid upgrade from v0.10.1 to v1
* Different definition of `def login_failure()`


johngian:
@akatsoulas r?

Verification steps:
* Checkout to ``master``
* Edit ``vendor-local/src/django-browserid/django_browserid/views.py``
 * To reproduce it change `` def post`` to return ``self.login_failure()`` always
* Try to login
* 500 server error should be raised
* Checkout to this PR 
* Try to login
 * You should be redirected to homepage

akatsoulas:
r+

PR #1247, opened 2015-09-18
Change the order of import.

akatsoulas:


johngian:
Imports look exactly the same like before:
16a694faecc3c5c611cc5bdc591a5b326954960a

johngian:
r+

PR #1250, opened 2015-09-23
[fix bug 746795] Add multiple curators per group.

akatsoulas:


akatsoulas:
@glogiotatidis @johngian r? pretty please?

mozillians/groups/api/v2.py
```diff
@@ -36,7 +36,7 @@ class GroupDetailedSerializer(GroupSerializer):
 
     class Meta:
         model = Group
-        fields = ('id', 'name', 'description', 'curator',
+        fields = ('id', 'name', 'description', 'curators',
```
glogiotatidis: API schema is not supposed to remove or alter keys. I suggest that you keep `curator` and you add a new key `curators` to list all curators of a group. Also don't forget to document, API is useless without docs!

mozillians/templates/groups/group.html
```diff
@@ -78,39 +78,44 @@ <h1 id="group_name">
...
       </ul><!-- .curated-group-details -->
 
-      {% if group.curator %}  {# curated group #}
+      {% set curators = group.curators.all() %}
+      {% if curators %}  {# curated group #}
         <div class="curator">
           <h3>
             {% if group.functional_area %}
               {{ _('Community Steward') }}
+            {% elif curators.count() > 1 %}
```
glogiotatidis: Pluralization needs special treatment in l10n. See http://jinja.pocoo.org/docs/dev/templates/#i18n

mozillians/groups/admin.py
```diff
@@ -195,6 +196,13 @@ def pending_member_count(self, obj):
         """Return number of members in group."""
         return obj.groupmembership_set.filter(status=GroupMembership.PENDING).count()
 
+    def get_curators(self, obj):
+        link = u"<a href='../../users/userprofile/%d/'>%s</a>"
```
glogiotatidis: use `reverse()`

mozillians/groups/admin.py
```diff
@@ -195,6 +196,13 @@ def pending_member_count(self, obj):
...
+        return mark_safe(', '.join(link % (tup[0], tup[1])
+                         for tup in [(user.id, user.full_name)
```
glogiotatidis: `user.full_name` should be escaped to avoid XSS since you `mark_safe`

glogiotatidis:
I did a very quick pass that should not be considered a full code review due to limited time. If time allows and this is still open I'll re-review later this week.

mozillians/groups/forms.py
```diff
@@ -31,7 +37,8 @@ def clean(self):
             cleaned_data['new_member_criteria'] = u''
         else:
             if not criteria:
-                msg = _(u'You must either specify the criteria or change the acceptance selection.')
+                msg = _(u'You must either specify the criteria or change the '
+                        'acceptance selection.')
```
johngian: Since PEP8 is not throwing any errors let's keep the initial version of this.

mozillians/groups/helpers.py
```diff
@@ -21,3 +21,23 @@ def slugify(s):
     if isinstance(s, str):
         s = unicode(s)
     return django_slugify(unidecode(s))
+
+
+@register.function
+def user_is_curator(group, userprofile):
+    """Chech if a user is curator in the specific group."""
```
johngian: /s/chech/check/g/

mozillians/groups/models.py
```diff
@@ -150,10 +151,8 @@ class Group(GroupBase):
     description = models.TextField(max_length=1024,
                                    verbose_name=_lazy(u'Description'),
                                    default='', blank=True)
-    curator = models.ForeignKey('users.UserProfile',
-                                blank=True, null=True,
-                                on_delete=models.SET_NULL,
-                                related_name='groups_curated')
+    curators = models.ManyToManyField('users.UserProfile',
+                                      related_name='groups_curated_new')
```
johngian: Why don't we just keep the same ``related_name``. I don't see any semantic meaning on ``groups_curated_new``.

mozillians/groups/tests/test_models.py
```diff
@@ -172,93 +172,93 @@ def test_pending_can_leave(self):
...
-    def test_has_member(self):
-        user = UserFactory.create()
-        group = GroupFactory.create()
-        ok_(not group.has_member(user.userprofile))
-        GroupMembership.objects.create(userprofile=user.userprofile, group=group,
-                                       status=GroupMembership.MEMBER)
-        ok_(group.has_member(user.userprofile))
-        group.remove_member(user.userprofile)
-        ok_(not group.has_member(user.userprofile))
+        def test_nonmember_cant_leave(self):
```
johngian: Please fix indentation. The following three tests need to go one indentation level up.

mozillians/groups/tests/test_models.py
```diff
@@ -172,93 +172,93 @@ def test_pending_can_leave(self):
...
+            group.add_member(user.userprofile)
+            ok_(not group.user_can_leave(user.userprofile))
+
+        def test_unique_name(self):
+            group = GroupFactory.create()
+            GroupAliasFactory.create(alias=group, name='bar')
+            group_2 = GroupFactory.build(name='bar')
+            self.assertRaises(ValidationError, group_2.clean)
+
+    class GroupTests(TestCase):
```
johngian: Same here. ``GroupTests`` class should go one indentation level up.

mozillians/groups/views.py
```diff
@@ -335,30 +340,40 @@ def group_add_edit(request, url=None):
...
     if form.is_valid():
         group = form.save()
-        # Ensure curator is in the group when it's created
-        if profile == group.curator and not group.has_member(profile):
+        for curator in form.cleaned_data['curators']:
+            group.curators.add(curator)
+            if not group.has_member(curator):
+                group.add_member(curator)
+        if created:
+            group.curators.add(profile)
```
johngian: Isn't this redundant after line 366. Also in case when ``created=True`` this will through an error if curator already added after line 366.

mozillians/templates/groups/group.html
```diff
@@ -78,39 +78,44 @@ <h1 id="group_name">
...
-                      {{ group.curator.display_name }}
+                    <a href="{{ url('phonebook:profile_view', curator.user.username) }}">
+                      {{ curator.display_name }}
                     </a>
                   </span>
                 </li>
                 <li>
-                  <a href="mailto:{{ group.curator.user.email }}">
+                  <a href="mailto:{{ curator.user.email }}">
                      <i class="icon-envelope-o"></i>
```
johngian: Please fix identation.

mozillians/templates/includes/search_result.html
```diff
@@ -1,7 +1,7 @@
 <div class="result">
   {% if group %}
     {% if is_curator %}
-      {% if user != profile.user %}
+      {% if user != profile.user and not user_is_curator(group, profile) %}
```
johngian: In previous version, this ``if case`` didn't check group's curator. Why do we need this now?

johngian:
This PR might need rebase and change in migrations after #1268

mozillians/groups/forms.py
akatsoulas: It does throw an error. It has 1 character more (100 instead of 99).

johngian:
Tests are failing.

mozillians/groups/api/v2.py
```diff
@@ -33,14 +33,18 @@ def get_url(self, obj):
 
 class GroupDetailedSerializer(GroupSerializer):
     members = GroupMemberSerializer(many=True, source='_members')
+    curator = serializers.SerializerMethodField('get_curator')
 
     class Meta:
         model = Group
-        fields = ('id', 'name', 'description', 'curator',
+        fields = ('id', 'name', 'description', 'curator', 'curators',
```
johngian: Please add a comment that ``curator`` field is for compatibility only and describe the difference of ``curator`` and ``curators``.

mozillians/groups/admin.py
```diff
@@ -208,6 +211,13 @@ def pending_terms_member_count(self, obj):
         """Return number of members in group who haven't accepted terms yet."""
         return obj.groupmembership_set.filter(status=GroupMembership.PENDING_TERMS).count()
 
+    def get_curators(self, obj):
+        url = u"<a href='%s'>%s</a>"
```
johngian: Instead of ``%s`` please use ``format``.

mozillians/groups/admin.py
```diff
@@ -208,6 +211,13 @@ def pending_terms_member_count(self, obj):
...
+        return mark_safe(', '.join([url % (reverse('admin:users_userprofile_change',
```
johngian: This is getting a bit obscure. Please calculate ``urls`` before return.

mozillians/groups/api/v2.py
```diff
@@ -66,7 +70,7 @@ class GroupFilter(django_filters.FilterSet):
 
     class Meta:
         model = Group
-        fields = ('name', 'functional_area', 'curator',
+        fields = ('name', 'functional_area', 'curators',
```
johngian: This breaks public API functionality.
* ``curator`` filter should remain the same
* ``curators`` should filter all curators

mozillians/groups/forms.py
```diff
@@ -22,6 +23,11 @@ def clean_sort(self):
 
 
 class GroupForm(happyforms.ModelForm):
+    # TODO: this probably will need some more filtering
+    curators = forms.ModelMultipleChoiceField(
+        queryset=UserProfile.objects.filter(is_vouched=True).exclude(full_name=''),
```
johngian: Isn't this initialized only once (first run)? I am not sure if that's initialized on form instatiation.

mozillians/groups/helpers.py
```diff
@@ -21,3 +21,23 @@ def slugify(s):
     if isinstance(s, str):
         s = unicode(s)
     return django_slugify(unidecode(s))
+
+
+@register.function
+def user_is_curator(group, userprofile):
+    """Check if a user is curator in the specific group."""
+    if userprofile in group.curators.all():
```
johngian: Nit: This could be one clean line eg:
```
return group.curators.filter(profile=profile).exists()
```

mozillians/groups/helpers.py
```diff
@@ -21,3 +21,23 @@ def slugify(s):
...
+
+@register.function
+def is_group_instance(obj):
+    """Check if the obj is of Group type."""
+
+    # Avoid circular dependencies
+    from mozillians.groups.models import Group
+
+    if isinstance(obj, Group):
+        return True
```
johngian: Same here:
```
return isinstance(obj, Group)
```

mozillians/groups/models.py
```diff
@@ -324,3 +323,10 @@ class Meta:
 
 class Skill(GroupBase):
     ALIAS_MODEL = SkillAlias
+
+    def user_can_leave(self, userprofile):
```
johngian: Is this a bug on master? Why do we need this now?

mozillians/groups/tests/test_models.py
```diff
@@ -219,11 +218,12 @@ def test_get_functional_areas(self):
...
     def test_deleted_curator_sets_null(self):
         user = UserFactory.create()
-        group = GroupFactory.create(curator=user.userprofile)
+        group = GroupFactory.create()
+        group.curators.add(user.userprofile)
 
         user.delete()
         group = Group.objects.get(id=group.id)
-        eq_(group.curator, None)
+        eq_(group.curators.all(), None)
```
johngian: Please be more explicit:
``eq_(group.curators.all().count(), 0)``

mozillians/groups/tests/test_tasks.py
```diff
@@ -36,7 +36,8 @@ def test_sending_pending_email(self):
         # was last sent, send the curator an email.  It should contain the count of
         # all pending memberships.
         curator = UserFactory.create()
-        group = GroupFactory.create(curator=curator.userprofile)
```
johngian: Instead of fixing all tests manually please amend this functionality in ``GroupFactory``. At the moment the expectancy is that the group has a curator.

mozillians/groups/views.py
```diff
@@ -221,14 +221,18 @@ def remove_member(request, url, user_pk):
         if profile_to_remove != this_userprofile:
             raise Http404()
 
-    # Curators cannot be removed, by anyone at all.
-    if group.curator == profile_to_remove:
+    # Curators cannot be removed, only by themselves and if there is another curator.
+    curators = group.curators.all()
+    if (profile_to_remove in curators and not curators.count() > 1 and
```
johngian: Please change this to:
`` if (profile_to_remove in curators and curators.count() <=1 and profile_to_remove != this_userprofile)``

mozillians/groups/views.py
```diff
@@ -382,26 +387,31 @@ def group_add_edit(request, url=None):
...
+        group = Group()
 
     form_class = SuperuserGroupForm if is_manager else GroupForm
 
-    form = form_class(request.POST or None, instance=group)
+    curators_ids = None
+    if url:
+        curators_ids = group.curators.all().values_list('id', flat=True)
+    form = form_class(request.POST or None, instance=group,
+                      initial={'curators': curators_ids})
```
johngian: Why are you using ``initial`` here? Shouldn't this be populated as ``ModelForm`` directly?

mozillians/groups/views.py
```diff
@@ -382,26 +387,31 @@ def group_add_edit(request, url=None):
...
     if form.is_valid():
         group = form.save()
-        # Ensure curator is in the group when it's created
-        if profile == group.curator and not group.has_member(profile):
-            group.add_member(profile)
+        for curator in form.cleaned_data['curators']:
+            group.curators.add(curator)
```
johngian: Maybe this functionality is better to be part of ``ModelForm.save()``.

mozillians/groups/forms.py
akatsoulas: Not it's populated every time you are using the form.

mozillians/groups/views.py
akatsoulas: Gladly but I don't see what's the difference in doing this. It won't make it any simpler or more readable. It's just a matter of coding style.

mozillians/groups/tests/test_tasks.py
akatsoulas: Personally, I prefer a simpler factory and I believe it's better to explicitly define what you are doing in each test, even if this requires a few lines of code more.

mozillians/groups/views.py
johngian: I don't have a strong opinion on that. It just came up while reviewing this, that it is more intuitive the way I describe. Feel free to choose, both are correct coding-wise.

mozillians/templates/includes/search_result.html
akatsoulas: Because if we don't include this now, a curator will be able to remove another curator.

mozillians/groups/views.py
akatsoulas: Just to pre-populate the form when editing a group.

mozillians/groups/models.py
akatsoulas: No, fortunately we don't have a bug. The reason we need this now and not before, is because we added the `curators` queryset in the `user_can_leave` method in the `GroupBase`. This has as a result, when we dealing with skills, to produce a 500 error since skills don't have a similar queryset.

johngian:
Creating a new group doesn't automatically add me as a curator nor as a group member.

akatsoulas:
@johngian pr updated

johngian:
Tests are failing.

johngian:
``GroupFactory`` needs some compatibility work. In the past ``groups`` created using factory had a curator assigned, now this is ``None``.

mozillians/groups/helpers.py
```diff
@@ -21,3 +21,18 @@ def slugify(s):
...
+    return group.curators.filter(user=userprofile).exists()
+
+
+@register.function
+def is_group_instance(obj):
+    """Check if the obj is of Group type."""
+
+    # Avoid circular dependencies
+    from mozillians.groups.models import Group
+    return isinstance(obj, Group)
```
johngian: Please use ``get_model`` instead.

johngian:
I am getting ``local variable 'created' referenced before assignment`` when removing the last curator.

akatsoulas:
Probably you just need to update your branch

johngian:
Groups can end up with no curator.

johngian:
Tested it locally:
* Creating a new group with multiple curators
* Removing curators
 
and works OK.

johngian:
r+wc

PR #1252, opened 2015-09-28
[fix bug 1198258] Fix 500 error when adding new group in admin.

akatsoulas:
@glogiotatidis r?

### DIFF
--- a/mozillians/groups/admin.py
+++ b/mozillians/groups/admin.py
@@ -170,11 +170,10 @@ class GroupAdmin(GroupBaseAdmin):
 
     fieldsets = (
         ('Group', {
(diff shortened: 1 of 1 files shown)


glogiotatidis:
Please explain what's happening and why this worked (?) before?

akatsoulas:
Yap, sorry for the omission!
I am not really sure when this regression was first introduced, but seems to be present for a while now.
The `merge_with` is declared as a field but the `add_form` does not have such a field and this produces a 500 error when adding a new group through the admin.

Basically, with this PR, I reversed the logic and removed completely the field. The field is conditionally introduced only if we need to edit a group.

glogiotatidis:
r+

akatsoulas:
@glogiotatidis thanks :beers:

PR #1253, opened 2015-09-29
[bug 1129848] Fix missing jquery library in admin.

akatsoulas:
@glogiotatidis quick r?


akatsoulas:
nevermind, merging it as is

PR #1256, opened 2015-10-14
[Fix bug 1214010] Fix DB constraint violations for region, country.

johngian:
* Check if duplicate region/country exists
* Deduplicate region/country and their references if required
* Workaround similar to https://github.com/mozilla/mozillians/pull/1178

### DIFF
--- a/mozillians/geo/lookup.py
+++ b/mozillians/geo/lookup.py
@@ -109,6 +109,27 @@ def result_to_country_region_city(result):
     return country, region, city
 
 
+def deduplicate_countries(country, dup_country):
+    """
+    Given 2 Country instances, deduplicate dup_country and it's references.
+    """
+    # Deduplicate geo data
+    regions = dup_country.region_set.all()
+    if regions.exists():
+        regions.update(country=country)
+
+    cities = dup_country.city_set.all()
+    if cities.exists():
+        cities.update(country=country)
+
+    # Deduplicate user data
+    users = dup_country.userprofile_set.all()
+    if users.exists():
+        users.update(geo_country=country)
+
+    dup_country.delete()
+
+
 def result_to_country(result):
     """
     Given one result from mapbox, converted to a dictionary keyed on 'type',
@@ -119,19 +140,47 @@ def result_to_country(result):
         mapbox_country = result['country']
         codes = dict((v, k) for k, v in product_details.get_regions('en-US').iteritems())
         code = codes.get(mapbox_country['name'], '')
-        country, created = Country.objects.get_or_create(
-            mapbox_id=mapbox_country['id'],
-            defaults=dict(
-                name=mapbox_country['name'],
-                code=code,
-            )
-        )
-        if not created:
-            # Update name if it's changed in mapbox
-            if country.name != mapbox_country['name']:
-                country.name = mapbox_country['name']
-                country.save()
+        lookup_args = {
+            'name': mapbox_country['name']
+        }
+        args = {
+            'mapbox_id': mapbox_country['id'],
+            'code': code
+        }
+
+        args.update(lookup_args)
+
+        query = Q(**lookup_args) | Q(mapbox_id=mapbox_country['id'])
+        country_qs = Country.objects.filter(query).distinct()
+
+        if country_qs.exists():
+            # Check if deduplication is required
+            if country_qs.count() == 2:
+                deduplicate_countries(country_qs[0], country_qs[1])
+
+            country_qs.update(**args)
+            country = country_qs[0]
+        else:
+            country = Country.objects.create(**args)
+
         return country
+
+
+def deduplicate_regions(region, dup_region):
+    """
+    Given 2 Country instances, deduplicate dup_country and it's references.
+    """
+    # Deduplicate geo data
+    cities = dup_region.city_set.all()
+    if cities.exists():
+        cities.update(region=region)
+
+    # Deduplicate user data
+    users = dup_region.userprofile_set.all()
+    if users.exists():
+        users.update(geo_region=region)
+
+    dup_region.delete()
 
 
 def result_to_region(result, country):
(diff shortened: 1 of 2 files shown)


johngian:
@akatsoulas r?

glogiotatidis:
Thanks for working on this @johngian :) Wouldn't it be useful to add a real db constrain and make this a data migration?

johngian:
We already have constraints in our models. The issue we are facing is that mapbox data are not consistent all the time and this causes violations of those constraints.

eg. We have in our DB a ``Country: name=Greece, mapbox_id=42`` that has been stored in the past and now mapbox returns for ``Country: name=Greece`` a different ``mapbox_id``. That's why we are adding this logic to update our data without duplicates.

We faced the same issue in the past for ``City`` and this fix seems to be working fine.

glogiotatidis:
thanks for explaining @johngian

mozillians/geo/lookup.py
```diff
@@ -119,21 +140,49 @@ def result_to_country(result):
         mapbox_country = result['country']
         codes = dict((v, k) for k, v in product_details.get_regions('en-US').iteritems())
         code = codes.get(mapbox_country['name'], '')
-        country, created = Country.objects.get_or_create(
+        lookup_args = dict(
```
akatsoulas: Please replace dict with `{}`. It's both faster and more pythonic (imo)

johngian: I agree. Only added it this way because of previous code.
/me changing it

mozillians/geo/lookup.py
```diff
@@ -119,21 +140,49 @@ def result_to_country(result):
...
+            name=mapbox_country['name']
+        )
+        args = dict(
```
akatsoulas: same here and everywhere else this might exist

mozillians/geo/lookup.py
```diff
@@ -119,21 +140,49 @@ def result_to_country(result):
...
         )
-        if not created:
-            # Update name if it's changed in mapbox
-            if country.name != mapbox_country['name']:
-                country.name = mapbox_country['name']
-                country.save()
+
+        args.update(lookup_args)
+
+        query = Q(**lookup_args) | Q(mapbox_id=mapbox_country['id'])
```
akatsoulas: No need for the `mapbox_id` query

mozillians/geo/lookup.py
```diff
@@ -119,21 +140,49 @@ def result_to_country(result):
...
+        country_qs = Country.objects.filter(query).distinct()
+
+        if country_qs.exists():
+            # Check if deduplication is required
+            if country_qs.count() == 2:
```
akatsoulas: Is there a possibility to have a number greater than 2?

johngian: In the same workaround for ``City`` we decided that this heuristic is good enough. I suggest we keep it the same here as well.

mozillians/geo/tests/test_lookup.py
```diff
@@ -106,6 +106,36 @@ def test_country_code_set(self):
         ok_(isinstance(country, Country))
         eq_(country.code, u'gr')
 
+    def test_deduplication_required(self):
```
akatsoulas: I am not sure if and why this test is needed. You are creating two countries with different names and different ids and then you query with a data set that will never exist in the database (the name of the one country with the id of the other)

mozillians/geo/tests/test_lookup.py
```diff
@@ -106,6 +106,36 @@ def test_country_code_set(self):
...
+        result_to_country(result)
+        lookup_args = {
+            'name': dup_country.name,
+        }
+
+        ok_(Country.objects.filter(**lookup_args).exists())
+        ok_(not Country.objects.filter(mapbox_id=dup_country.mapbox_id).exists())
+
+    def test_deduplicate_countries(self):
+        countries = CountryFactory.create_batch(2)
```
akatsoulas: Instead of creating different countries, I would suggest to create two countries with exactly the same name and different IDs.

mozillians/geo/tests/test_lookup.py
```diff
@@ -140,6 +170,37 @@ def test_update_name(self):
         region = Region.objects.get(pk=region.pk)
         eq_(new_name, region.name)
 
+    def test_deduplication_required(self):
```
akatsoulas: The above comments apply here too.

johngian:
@akatsoulas PR updated

mozillians/geo/tests/test_lookup.py
```diff
@@ -106,6 +107,40 @@ def test_country_code_set(self):
...
+        country_qs = Country.objects.filter(name=country.name)
+        eq_(country_qs.count(), 1)
+        eq_(country_qs[0].mapbox_id, '24')
+        ok_(not Country.objects.filter(mapbox_id='42').exists())
+
+        userprofile = UserProfile.objects.get(pk=user.userprofile.id)
+        eq_(userprofile.geo_country.name, country.name)
+        eq_(userprofile.geo_country.mapbox_id, '24')
+
+    def test_deduplicate_countries(self):
```
akatsoulas: I still believe this test is not needed. The test above does exactly what you are trying to test here. That being said, if you do decide to keep it, you need to create two countries with the same name and different IDs. It won't make any difference regarding the function but it will be much clearer regarding the context of the test

akatsoulas: Probably, it would be better to mock the function and just test that the conditions are met (there are dup countries) and the function is just called.

akatsoulas: The same comments apply to the region too

akatsoulas:
r+wc :beers:

akatsoulas:
Please add docstrings to the tests documenting why are performing this kind of tests.

PR #1260, opened 2015-10-20
[Fix bug 1216291] Fix photo field definition in users api serializer.

johngian:


johngian:
Verified it locally:

* Checkout to ``master``
* GET ``127.0.0.1:8000/api/v2/users/<user_id>`` with users's photo field permissions higher than the permissions of the API key used
* traceback

* Checkout to this PR
* GET
* no traceback

johngian:
@akatsoulas r?

akatsoulas:
Reproduced this locally with the steps above. r+ :beer:

PR #1261, opened 2015-10-27
Add `pt-PT` in supported locales.

johngian:


akatsoulas:
code looks good but travis is not very happy
r+wc

johngian:
The database was superbusy apparently and went away :rocket:

PR #1262, opened 2015-10-28
[fix bug 1218776] Change the max length in groups models.

akatsoulas:
* Changed description to 1024 chars.
* Changed new_member_criteria to 1024 chars.


comzeradd:
Code looks good. r+

PR #1265, opened 2015-11-24
Vouch mozilla addresses

akatsoulas:
@johngian or @glogiotatidis r?

### DIFF
--- a/mozillians/users/migrations/0003_auto_20151124_0647.py
+++ b/mozillians/users/migrations/0003_auto_20151124_0647.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+import operator
+from datetime import datetime
+
+from django.db import models, migrations
+from django.conf import settings
+
+
+def vouch_mozilla_alternate_emails(apps, schema_editor):
+    """Vouch all users with a mozilla* email address as alternate."""
+    ExternalAccount = apps.get_model('users', 'ExternalAccount')
+    UserProfile = apps.get_model('users', 'UserProfile')
+
+    email_query = reduce(operator.or_,
+                         (models.Q(identifier__icontains=item)
+                          for item in settings.AUTO_VOUCH_DOMAINS))
+    user_ids = (ExternalAccount.objects.filter(type='EMAIL').filter(email_query)
+                .values_list('user_id', flat=True).distinct())
+    users = UserProfile.objects.filter(id__in=user_ids)
+
+    q_args = {
+        'autovouch': True,
+        'description': settings.AUTO_VOUCH_REASON
+    }
+    for user in users:
+        if not user.vouches_received.filter(**q_args).exists():
+            if not user.vouches_received.all().count() >= settings.VOUCH_COUNT_LIMIT:
+                user.vouches_received.create(
+                    voucher=None,
+                    date=datetime.now(),
+                    description=settings.AUTO_VOUCH_REASON,
+                    autovouch=True
+                )
+                vouches = user.vouches_received.all().count()
+                user.is_vouched = vouches > 0
+                user.can_vouch = vouches > settings.CAN_VOUCH_THRESHOLD
+                user.save()
+
+
+def unvouch_mozilla_alternate_emails(apps, schema_editor):
+    """Vouches cannot be removed, do nothing please."""
+    pass
+
+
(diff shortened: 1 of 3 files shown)


mozillians/users/migrations/0003_auto_20151124_0647.py
```diff
@@ -0,0 +1,46 @@
...
+from django.db import models, migrations
+from django.conf import settings
+
+
+def vouch_mozilla_alternate_emails(apps, schema_editor):
+    """Vouch all users with a mozilla* email address as alternate."""
+    ExternalAccount = apps.get_model('users', 'ExternalAccount')
+    UserProfile = apps.get_model('users', 'UserProfile')
+
+    user_ids = (ExternalAccount.objects.filter(type='EMAIL', identifier__icontains='@mozilla')
```
johngian: Please filter explicitly using ``settings.AUTO_VOUCH_DOMAINS``.

mozillians/users/models.py
```diff
@@ -488,12 +488,16 @@ def vouch(self, vouched_by, description='', autovouch=False):
...
-
-        if any(email.endswith('@' + x) for x in settings.AUTO_VOUCH_DOMAINS):
-            if not self.vouches_received.filter(
-                    description=settings.AUTO_VOUCH_REASON, autovouch=True).exists():
-                self.vouch(None, settings.AUTO_VOUCH_REASON, autovouch=True)
+        emails = [acc.identifier for acc in
+                  ExternalAccount.objects.filter(user=self, type=ExternalAccount.TYPE_EMAIL)]
+        emails.append(self.user.email)
+
+        for email in emails:
```
johngian: Instead of iterating ``for email in emails`` and then ``break`` you could do something like this:

```
emails.filter(lambda x: any(email.endswith('@', x) for x in settings.AUTO_VOUCH_DOMAINS))
if emails:
    etc
```

mozillians/users/models.py
```diff
@@ -865,6 +869,15 @@ def __unicode__(self):
         return self.type
 
 
+@receiver(dbsignals.post_save, sender=ExternalAccount, dispatch_uid='add_employee_vouch_sig')
+def add_employee_vouch(sender, instance, **kwargs):
+    """Add a vouch if an alternate email address is a mozilla* address."""
+
+    if kwargs.get('raw') or not instance.type == ExternalAccount.TYPE_EMAIL:
```
johngian: Why do we need ``kwargs.get('raw')``?

mozillians/users/tests/test_models.py
```diff
@@ -125,6 +125,37 @@ def test_vouch_can_vouch_gets_updated(self):
         unvouched = User.objects.get(pk=unvouched.id)
         eq_(unvouched.userprofile.can_vouch, True)
 
+    def test_vouch_alternate_mozilla_address(self):
+        user = UserFactory.create(vouched=False)
+        user.userprofile.externalaccount_set.create(type=ExternalAccount.TYPE_EMAIL,
+                                                    identifier='test@mozilla.com')
+        eq_(Vouch.objects.filter(vouchee=user.userprofile).count(), 1)
```
johngian: Please also check that Vouch is ``autovouch`` and reason is ``AUTOVOUCH_REASON``.

johngian: This also has other occurences.

johngian:
Please add a test that checks that everythings works OK when a user changes an alternate ``@mozilla`` email to primary.

mozillians/users/models.py
akatsoulas: `emails` is a list and not a queryset. The reason for this is because we need both the primary and the alternate emails.

akatsoulas:
@johngian pr updated

mozillians/users/models.py
johngian: I meant https://docs.python.org/2/library/functions.html#filter

johngian:
Tested it locally:

* Adding a new external ``@mozilla`` email works
* Datamigration is not working because signals dont update ``is_vouched``.

johngian:
Tested it locally:

* Adding new ``*mozilla`` alternate email adds automatic vouch
* Datamigration works as expected

johngian:
r+

johngian:
Heads up: Please check push logs on ``dev`` to check for any errors on ``datamigration``.

PR #1266, opened 2015-11-24
Don't allow users to join group without accepting its terms.

johngian:


johngian:
@akatsoulas r?

akatsoulas:
Reminder: if PR #1250 is merged first, then the migration here needs update and vice versa.

mozillians/groups/urls.py
```diff
@@ -12,6 +12,12 @@
     url('^groups/add/$', 'views.group_add_edit', name='group_add'),
     url('^group/(?P<url>[-\w]+)/edit/$', 'views.group_add_edit', name='group_edit'),
     url('^group/(?P<url>[-\w]+)/delete/$', 'views.group_delete', name='group_delete'),
+    url('^group/(?P<url>[-\w]+)/terms/$', 'views.group_terms',
+        name='group_terms'),
+    url('^group/(?P<url>[-\w]+)/terms/accept$', 'views.accept_group_terms',
+        name='accept_group_terms'),
+    url('^group/(?P<url>[-\w]+)/terms/deny$', 'views.deny_group_terms',
+        name='deny_group_terms'),
```
akatsoulas: Can't you merge accept and deny in a single view?

johngian: I don't see why it's better to check ``if foo == 'accept' elif foo == 'deny'``. This way is much cleaner.

johngian:
@akatsoulas PR updated after #1268

mozillians/groups/urls.py
akatsoulas: I would suggest to use a single url with a single view. Less fragmentation, thus easier to follow what the code is doing.

mozillians/groups/views.py
```diff
@@ -265,6 +272,42 @@ def edit(request, url, alias_model, template):
...
+def group_terms(request, url):
+    """Group terms page."""
+    group = get_object_or_404(Group, url=url)
+    if not group.terms:
+        return redirect(reverse('groups:show_group', args=[group.url]))
+    return render(request, 'groups/terms.html', {'group': group})
+
+
+def accept_group_terms(request, url):
+    """Accept group terms."""
```
akatsoulas: Please be a bit more descriptive on what the view is doing.

akatsoulas:
Probably you need to update the `has_member` method in the `groups/models.py` to include the `has_terms` flag.

akatsoulas:
What happens regarding the acceptance of terms, if exist, for the creators of a group?

mozillians/groups/models.py
```diff
@@ -196,6 +197,7 @@ class Group(GroupBase):
                    u'curator a reminder')
     )
 
+    terms = models.TextField(default='', verbose_name=_('Terms'))
```
akatsoulas: `blank=True`?

akatsoulas: Also this is missing from the form in edit groups.

johngian:
PR updated
* Group creator accepts terms directly when creating a group
* I added some more tests

mozillians/templates/groups/terms.html
```diff
@@ -0,0 +1,24 @@
+{% extends "base.html" %}
+
+{% block page_title %}{{ _('Group terms') }}{% endblock %}
```
akatsoulas: s/Group terms/Review Terms ?

akatsoulas:
You are missing a migration here. Please amend the changes in the `0006` migration.

mozillians/groups/forms.py
```diff
@@ -40,7 +41,7 @@ class Meta:
         model = Group
         fields = ['name', 'description', 'irc_channel',
                   'website', 'wiki', 'accepting_new_members',
-                  'new_member_criteria']
+                  'new_member_criteria', 'terms']
```
akatsoulas: Why is this needed here?

mozillians/groups/forms.py
```diff
@@ -72,3 +74,17 @@ def clean_filtr(self):
         if self.cleaned_data['filtr'] == '':
             return 'all'
         return self.cleaned_data['filtr']
+
+
+class GroupMembershipForm(forms.ModelForm):
+    class Meta:
+        model = GroupMembership
+        fields = ['terms_accepted', ]
```
akatsoulas: Please remove the comma

mozillians/groups/forms.py
johngian: This is GroupForm. ``git diff`` is misleading

mozillians/groups/views.py
```diff
@@ -265,6 +273,32 @@ def edit(request, url, alias_model, template):
     return render(request, alias_model, template)
 
 
+def group_terms(request, url):
```
akatsoulas: I would go for something more descriptive. Like review_group_terms or plain review_terms

mozillians/groups/views.py
```diff
@@ -354,6 +388,10 @@ def group_add_edit(request, url=None):
         # Ensure curator is in the group when it's created
         if profile == group.curator and not group.has_member(profile):
             group.add_member(profile)
+            if group.terms:
```
akatsoulas: This is rather interesting. Although it makes sense to auto-accept the terms for a curator, at the same time you are accepting terms that you have never reviewed. Moreover if you factor in PR #1250, the you have curators adding new curators where the new curators have the terms accepted without even logging into the portal. I don't have a clear suggestion here, I am just flagging this.  Thoughts? cc @ppapadeas

akatsoulas:
Should this functionality be available only for groups that require approval to join or every group?

akatsoulas:
Right after a request to join a group, I am redirected to a page to accept/decline the group terms. This page should only be displayed right after the curator has accepted my request.

mozillians/groups/views.py
johngian: I suggest to treat curators the same way with the rest of the users. And if another UX comes we can ammend this.

akatsoulas: Well if this is the case, then we should not auto-accept the terms for the curators as we do here. We should block their curator status until they accept the terms. The only exception should be the creator of the group.

johngian:
* This functionality should be required for every group with terms not only those requiring request.

johngian:
```
Right after a request to join a group, I am redirected to a page to accept/decline the group terms. This page should only be displayed right after the curator has accepted my request.
```

This was the initial implementation but I changed it. It makes sense though to be that way.

mozillians/groups/views.py
ppapadeas: When adding a curator it is obvious that this is a trusted person (already under multiple NDAs). Please proceed with auto adding to the group and auto accepting the legal text.

johngian:
PR updated:
* Code styling changes
* ``Review terms`` page is only shown to ``GroupMembership.MEMBER``.

mozillians/groups/forms.py
```diff
@@ -72,3 +74,17 @@ def clean_filtr(self):
...
+
+class GroupMembershipForm(forms.ModelForm):
+    class Meta:
+        model = GroupMembership
+        fields = ['terms_accepted']
+        widgets = {
+            'terms_accepted': RadioSelect(
+                choices=[
+                    (True, _('I accept the group terms.')),
+                    (False, _("I don't accept the group terms."))
```
akatsoulas: Maybe something like `I accept these terms` and `I don't accept these terms` is a bit nicer.

mozillians/groups/migrations/0007_auto_20151127_0439.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
akatsoulas: Please squash this migration together with `0006`

mozillians/groups/models.py
```diff
@@ -291,9 +293,12 @@ def remove_member(self, userprofile, send_email=True):
     def has_member(self, userprofile):
         """
         Return True if this user is in this group with status MEMBER.
+        If group has `terms` exclude users that haven't accepted the terms.
         """
-        return self.groupmembership_set.filter(userprofile=userprofile,
-                                               status=GroupMembership.MEMBER).exists()
+        qs = self.groupmembership_set.filter(userprofile=userprofile, status=GroupMembership.MEMBER)
```
akatsoulas: I believe this is error prone and will lead to future mistakes/bugs. The membership status should remain `PENDING` until the user has accepted the terms, if exist. The flow should be the following:

* status == MEMBER if there are no terms and users can freely join a group
* status == PENDING if a user can join a group by request and there are no terms
  ** status == MEMBER when the user is accepted
* status == PENDING if a user can join a group by request and there terms
  ** status == MEMBER when the user is accepted *and* has accepted the terms

johngian:
@akatsoulas PR updated
* Replaced ``terms_accepted`` with a new state in ``GroupMembership.status``
* Various minor fixes

mozillians/groups/models.py
```diff
@@ -123,15 +123,17 @@ class GroupMembership(models.Model):
     # Possible membership statuses:
     MEMBER = u'member'
     PENDING = u'pending'  # Has requested to join group, not a member yet
+    TERMS_PENDING = u'terms_pending'
```
akatsoulas: Please add a small comment, similar to the line above.

mozillians/groups/models.py
```diff
@@ -260,13 +263,18 @@ def add_member(self, userprofile, status=GroupMembership.MEMBER):
                 # Status changed
                 old_status = membership.status
                 membership.status = status
-                if (old_status, status) == (GroupMembership.PENDING, GroupMembership.MEMBER):
+                statuses = [(GroupMembership.PENDING, GroupMembership.MEMBER),
+                            (GroupMembership.PENDING, GroupMembership.TERMS_PENDING),
+                            (GroupMembership.TERMS_PENDING, GroupMembership.MEMBER)]
```
akatsoulas: I would suggest to add a few comments explaining the various combinations of statuses that you have here.

mozillians/groups/models.py
```diff
@@ -260,13 +263,18 @@ def add_member(self, userprofile, status=GroupMembership.MEMBER):
...
+                if (old_status, status) in statuses:
                     # Request accepted
```
akatsoulas: Update this comment please. This isn't true any more.

mozillians/groups/models.py
```diff
@@ -260,13 +263,18 @@ def add_member(self, userprofile, status=GroupMembership.MEMBER):
...
-                        # Group is functional area, we want to sent this update to Basket.
-                        update_basket_task.delay(userprofile.id)
-                    email_membership_change.delay(self.pk, userprofile.user.pk, old_status, status)
+                    if membership.status in [GroupMembership.PENDING, GroupMembership.MEMBER]:
+                        if self.functional_area:
+                            # Group is functional area, we want to sent this update to Basket.
+                            update_basket_task.delay(userprofile.id)
+                        email_membership_change.delay(self.pk, userprofile.user.pk,
+                                                      old_status, status)
                 # else? never demote people from full member to requested, that doesn't make sense
```
akatsoulas: Probably we need to remove this, since a user can be demoted (eg expiration of terms)

mozillians/groups/models.py
```diff
@@ -260,13 +263,18 @@ def add_member(self, userprofile, status=GroupMembership.MEMBER):
...
+                            (GroupMembership.PENDING, GroupMembership.TERMS_PENDING),
+                            (GroupMembership.TERMS_PENDING, GroupMembership.MEMBER)]
+                if (old_status, status) in statuses:
                     # Request accepted
                     membership.save()
-                    if self.functional_area:
-                        # Group is functional area, we want to sent this update to Basket.
-                        update_basket_task.delay(userprofile.id)
-                    email_membership_change.delay(self.pk, userprofile.user.pk, old_status, status)
+                    if membership.status in [GroupMembership.PENDING, GroupMembership.MEMBER]:
```
akatsoulas: What about the other statuses? Shouldn't a user get notified when for example is accepted to a group but hasn't yet accepted the terms?

mozillians/groups/tests/test_views/test_misc.py
```diff
@@ -181,6 +181,31 @@ def test_basic_group_creation_as_non_manager(self):
         eq_(data['website'] + '/', group.website)
         eq_(data['wiki'] + '/', group.wiki)
 
+    def test_basic_group_creation_with_terms(self):
+        # Managers have access to all group parameters when creating a group
```
akatsoulas: Probably this comment is irrelevant

mozillians/groups/tests/test_views/test_misc.py
```diff
@@ -181,6 +181,31 @@ def test_basic_group_creation_as_non_manager(self):
...
+            # 'functional_area' not checked
+        }
+        with self.login(user) as client:
+            response = client.post(url, data=data, follow=False)
+        eq_(302, response.status_code)
+        group = GroupAlias.objects.get(name=data['name']).alias
+        eq_(group.terms, 'Example terms')
+        membership = GroupMembership.objects.get(group=group, userprofile=user.userprofile)
+        ok_(membership.status, GroupMembership.TERMS_PENDING)
+
```
akatsoulas: Please add one more test for a group only joined `by request` and check how the status is changed in each step.

mozillians/groups/urls.py
```diff
@@ -12,6 +12,8 @@
     url('^groups/add/$', 'views.group_add_edit', name='group_add'),
     url('^group/(?P<url>[-\w]+)/edit/$', 'views.group_add_edit', name='group_edit'),
     url('^group/(?P<url>[-\w]+)/delete/$', 'views.group_delete', name='group_delete'),
+    url('^group/(?P<url>[-\w]+)/terms/$', 'views.review_terms',
+        name='review_terms'),
```
akatsoulas: This should fit in one line

mozillians/groups/views.py
```diff
@@ -256,7 +266,10 @@ def confirm_member(request, url, user_pk):
         if membership.status == GroupMembership.MEMBER:
             messages.error(request, _('This user is already a member of this group.'))
         else:
-            group.add_member(profile)
+            status = GroupMembership.MEMBER
+            if group.terms:
```
akatsoulas: Probably if you use `eiif else` the flow could be more clear.

mozillians/groups/views.py
```diff
@@ -265,6 +278,33 @@ def edit(request, url, alias_model, template):
...
+        return redirect(reverse('groups:show_group', args=[group.url]))
+
+    membership = get_object_or_404(GroupMembership, group=group,
+                                   userprofile=request.user.userprofile,
+                                   status=GroupMembership.TERMS_PENDING)
+
+    membership_form = TermsReviewForm(request.POST or None)
+    if membership_form.is_valid():
+        terms_accepted = membership_form.cleaned_data['terms_accepted']
+        if terms_accepted == 'True':
```
akatsoulas: Can't you combine this to:
```python
`if membership_form.cleaned_data['terms_accepted':
     # add membership here
else:
     # delete membership
```

mozillians/groups/forms.py
akatsoulas: Please add a placeholder for this field. Either as a widget attribute or in the html

mozillians/templates/groups/add_edit.html
```diff
@@ -68,6 +68,10 @@
       {{ mozillians_field(form.accepting_new_members) }}
     </fieldset>
 
+    <fieldset>
+      {{ mozillians_field(form.terms) }}
```
akatsoulas: This should be below the `form.new_member_criteria`

mozillians/templates/groups/add_edit.html
```diff
@@ -68,6 +68,10 @@
...
+    </fieldset>
+
     <fieldset id="id_new_member_criteria_fieldset">
       {{ mozillians_field(form.new_member_criteria, required=True) }}
```
akatsoulas: When this field exists, is not aligned with the `terms` field.

akatsoulas:
You should update the `pending_member_count` in admin.py to include members that haven't yet accepted the terms of a group.

akatsoulas:
Probably, the `get_annotated_groups` method in users/models.py needs a small update too.

mozillians/groups/models.py
johngian: Opening a new bug to track this, since this is out of the scope of this bug.

mozillians/groups/views.py
johngian: Group terms and else refer to the same case but with a minor change (``status``). If we use ``elif group.terms`` we are going to duplicate ``add_member`` code in ``elif/else``.

akatsoulas: :+1:

mozillians/templates/groups/add_edit.html
johngian: Opening a new bug for that since this is not a regression. This is already a bug in master.

mozillians/groups/views.py
johngian: :rabbit:

johngian:
@akatsoulas PR updated

mozillians/groups/admin.py
```diff
@@ -162,15 +162,16 @@ class GroupAdmin(GroupBaseAdmin):
     inlines = [GroupAliasInline]
     list_display = ['name', 'curator', 'functional_area', 'accepting_new_members',
                     'members_can_leave', 'visible', 'total_member_count', 'full_member_count',
-                    'pending_member_count']
+                    'pending_member_count', 'terms_pending_member_count']
```
akatsoulas: s/tems_pending_member_count/pending_terms_member_count ?

mozillians/groups/admin.py
```diff
@@ -199,9 +201,13 @@ def full_member_count(self, obj):
         return obj.groupmembership_set.filter(status=GroupMembership.MEMBER).count()
 
     def pending_member_count(self, obj):
-        """Return number of members in group."""
+        """Return number of pending members in group."""
         return obj.groupmembership_set.filter(status=GroupMembership.PENDING).count()
 
+    def terms_pending_member_count(self, obj):
+        """Return number of members in group who haven't accepted terms yet."""
+        return obj.groupmembership_set.filter(status=GroupMembership.TERMS_PENDING).count()
```
akatsoulas: This probably holds true for this too (PENDING_TERMS). Sorry for not catching this earlier.

akatsoulas:
r+wc :beers:

PR #1267, opened 2015-11-24
[fix bug 1226779] Upgrade django to version 1.7.11.

akatsoulas:
@johngian r?


johngian:
Code looks OK. Did some casual surfing and the upgrade doesn't look like its breaking any functionality.

r+

johngian:
Code looks OK. Did some casual surfing and the upgrade doesn't look like its breaking any functionality.

r+

PR #1268, opened 2015-11-26
[Fix bug 1207956] Remove `invalidate` flag.

johngian:
* `invalidate` flag is not useful any more
* We are going to use `terms_accepted` and `status` combination


akatsoulas:
r+

PR #1274, opened 2015-12-02
Invalid memberships after time period.

johngian:


mozillians/static/mozillians/js/group-edit.js
```diff
@@ -10,11 +14,25 @@ $(function(){
         }
     }
 
+    function checkMembershipInvalidation() {
```
akatsoulas: s/check/toggle

akatsoulas:
The user should be able to add an invalidation period only when there are terms in the group. I haven't tested this but it seems that the logic is missing for this part.

akatsoulas:
Well we are not there yet but before you merge (or at any point) squash these commits. This is a single bug functionality and should be delivered in a single commit.

mozillians/groups/forms.py
```diff
@@ -22,6 +22,12 @@ def clean_sort(self):
 
 
 class GroupForm(happyforms.ModelForm):
+    invalidation_days = forms.IntegerField(
+        widget=forms.NumberInput(attrs={'placeholder': 'days'}),
+        min_value=0,
```
akatsoulas: This shouldn't be at least one? If it is `0` it beats the whole purpose of this functionality. Probably, this should be checked in the task too.

mozillians/static/mozillians/js/group-edit.js
johngian: I prefer ``check..`` to be compatible with the rest of the ``functions`` in the same file.

akatsoulas:
Please keep in mind that if we merge PR #1250 first, you will need to update the migration here and vice versa.

mozillians/groups/models.py
```diff
@@ -135,6 +135,7 @@ class GroupMembership(models.Model):
     group = models.ForeignKey('groups.Group', db_index=True)
     status = models.CharField(choices=MEMBERSHIP_STATUS_CHOICES, max_length=15)
     date_joined = models.DateTimeField(null=True, blank=True)
+    updated_on = models.DateTimeField(auto_now=True, null=True)
```
akatsoulas: I would suggest to save this in `save` method, in case we ever need to edit this trough admin.

mozillians/groups/tasks.py
```diff
@@ -133,3 +135,27 @@ def member_removed_email(group_pk, user_pk):
...
+
+
+@periodic_task(run_every=timedelta(hours=24))
+def invalidate_group_membership():
+    """
+    For groups with defined `invalidation_days` we need to invalidate
+    user membership after timedelta.
+    """
+    Group = get_model('groups', 'Group')
+    GroupMembership = get_model('groups', 'GroupMembership')
```
akatsoulas: Please use a proper import statement here.

mozillians/groups/models.py
johngian: I suggest to not be able to edit this if we need to rely on the last ``update_on`` timestamp.

akatsoulas:
Tests are failing

akatsoulas:
Please add the `invalidation` field in the admin interface.

akatsoulas:
This PR needs a rebase and an update on the migration.

akatsoulas:
What happens in the case where the group has no more members due to the invalidation process? Should the group be deleted?

akatsoulas:
Apart from the last comment, the functionality seems to be working fine. In the case where a group with no members after the invalidation process should be deleted, just add a small test for this.
r+wc :100:

johngian:
@akatsoulas PR updated
* Updated migrations
* Updated comments
* Rebased
* Groups without members are deleted automatically every day

PR #1275, opened 2015-12-04
[Fix bug 1227479] Hide multiple curators profile details

comzeradd:


comzeradd:
@akatsoulas r?

mozillians/static/mozillians/css/main.less
```diff
@@ -2197,25 +2197,32 @@ body#group-show {
...
+            .curator {
+                min-width: 70px;
+                min-height: 70px;
+                display: inline-block;
+                .avatar {
+                    .rounded-image;
+                    width: 60px;
+                    height: 60px;
+                }
+                .details-not {
```
johngian: Instead of ``.details-not`` please rename this class to ``hide-details``.

mozillians/templates/groups/group.html
```diff
@@ -91,29 +91,31 @@ <h1 id="group_name">
...
-                      {{ curator.display_name }}
+            <div class="curator">
+              <div class="avatar">
+                <a href="{{ url('phonebook:profile_view', curator.user.username) }}">
+                  <img class="profile-photo"
+                       src="{{ curator.get_photo_url('70x70') }}"
+                       alt="{{ _('Profile Photo') }}">
+                </a>
+              </div>
+              <div class="details {% if curators|count > 3 %}details-not{% endif %}">
```
johngian: Why instead of adding ``details-not`` (meaning ``display: none``) don't you just skip rendering of details when we have more than 3 curators?

mozillians/static/mozillians/css/main.less
johngian: I am not sure if we should tackle ``details`` hiding with css or on render.

johngian:
On ``320x480`` and ``360x640`` mobile views email is not rendered correctly when we have less than 3 curators.

comzeradd:
@johngian PR updated

mozillians/templates/groups/group.html
```diff
@@ -91,32 +91,36 @@ <h1 id="group_name">
...
-            </div><!-- details -->
+            <div class="curator">
+              <div class="avatar">
+                <a href="{{ url('phonebook:profile_view', curator.user.username) }}">
+                  <img class="profile-photo"
+                       src="{{ curator.get_photo_url('70x70') }}"
+                       alt="{{ _('Profile Photo') }}">
+                </a>
+              </div>
+              {% if curators|count < 4 %}
```
johngian: This way on each loop we calculate count. Let's define this outside the loop (calculate only once) and use it directly as eg ``curators_count``.

comzeradd:
@johngian Thanks for the tip. PR updated.

mozillians/templates/groups/group.html
```diff
@@ -90,33 +90,40 @@ <h1 id="group_name">
               {{ _('Curator') }}
             {% endif %}
           </h3>
+          {% if curators|count < 4 %}
+            {% set curators_details = 'true' %}
```
johngian: Just use a boolean instead of string 'true'.
A more elegant way to define this is:
`` {% set curators_details = curators.count() < 4 %}``

Please check if snippet needs any update to work but this is the general idea.

johngian: Also here curators.count > 4 curators_details is undefined.

comzeradd:
PR updated. I tested it locally and it works as expected.

mozillians/templates/groups/group.html
```diff
@@ -90,33 +90,38 @@ <h1 id="group_name">
               {{ _('Curator') }}
             {% endif %}
           </h3>
+          {% set curators_details = curators.count() < 4 %}
```
johngian: Please rename this var to ``show_curators_details`` in order to be more explicit what's going on.

mozillians/templates/groups/group.html
```diff
@@ -90,33 +90,38 @@ <h1 id="group_name">
...
-            </div><!-- details -->
+            <div class="curator">
+              <div class="avatar">
+                <a href="{{ url('phonebook:profile_view', curator.user.username) }}">
+                  <img class="profile-photo"
+                       src="{{ curator.get_photo_url('70x70') }}"
+                       alt="{{ _('Profile Photo') }}">
+                </a>
+              </div>
+              {% if curators_details %}
```
akatsoulas: I would suggest to add a tooltip, when we are not showing profile details, so that a user can see the `display_name` on mouse over.

comzeradd:
PR updated. I amended the commit to add @akatsoulas nice idea for adding a tooltip when we don't show curator's details.

johngian:
@comzeradd I can't see the tooltip locally. Is the PR up-to-date?

comzeradd:
@johngian for consistency I enabled jquery-ui tooltip site-wide. Keep in mind that the tooltip is displayed only when curator details are omitted.

johngian:
I did a cross-check across the site and some tooltips show redundant information and don't look really good. eg.

* In group list
 * Each group tab shows the name and hover a similar UI element shows the name once again
* In base
 * Sign in button has a sign-in text and on hover it show "sign-in" once again

I don't think that it looks good. Let's restrict this only when we need it. For now this is only group curators in group view page and only if there are no curator details.

johngian:
I did a cross-check across the site and I don't really like the way some tooltips get rendered. Eg.

* In group list
 * Each group tab shows the name and hover a similar UI element shows the name once again
* In base
 * Sign in button has a sign-in text and on hover it show "sign-in" once again

I know that the same think happens with the default title on hover but jquery-ui tooltip seems a bit too much for some cases.

I don't have any strong feelings on that. @akatsoulas @nikos any ideas on what is better?

comzeradd:
The point of a tooltip it to provide additional information. So I would suggest we remove the title element from these two cases you mention. It doesn't offer anything.

Long term, we could also use [boostrap's tooltip](http://getbootstrap.com/javascript/#tooltips), since we already use bootstrap, and completely remove jquery-ui if it's not used anywhere else.

johngian:
OK, for me our aim should be to not degrade the current UI just for a minor change. If we want to keep this tooltip please review the rest of the occurrences and omit the ``title``. Else let's restrict to the scope of the current PR and open a bug to tackle the rest of the tooltips in the future.

akatsoulas:
I believe that we should aim for code maintainability, thus I am in favor of just adding the tooltip exactly and only where we need it. Removing just the title in a couple of cases is not a good idea in the long run. By revisiting this in a few months, nobody is going to remember why this was done in the first place. Moreover, it's not helpful either for contributors or anybody else that is working/going to work on the code-base.

comzeradd:
I amended the commit to use the tooltip just for the multiple curators fix for the moment. I'll open a bug to review tooltips and jquery-ui in general for the whole website.

johngian:
@comzeradd 
Tested locally and it works fine. Also code looks OK.
Please rename ``groups.js`` to ``group_view.js`` following our current static file paradigm. Let's make sure that all references are also updated.

r+wc

PR #1276, opened 2015-12-09
[bug 746795] Make the curators field in the form required.

akatsoulas:


akatsoulas:
@johngian r?

johngian:
Tested it locally and field is required. Code looks ok. r+

PR #1286, opened 2016-01-10
[Fix 1232978] Removed extra trailing slash (/) from Telegram URL to fix the links

shinenelson:
the `flake8 mozillians` command threw loads of errors when I ran it right after forking and cloning. I don't know if this is specific to my development environment. My commit is to just change a line. I don't think that would introduce a lot of errors. I'm happy to change something if necessary.


comzeradd:
Hey,

Thanks for the Pull Request.

Can you reopen it against the master branch instead of next? You should also rebase it with master so it doesn't push all these commits along.

PR #1289, opened 2016-01-15
[Fix bug 1227293] Do not cache API v2 responses.

johngian:


johngian:
@akatsoulas
PR updated using ``never_cache`` decorator.

akatsoulas:
code looks good, r+

PR #1294, opened 2016-01-26
[Fix bug 1242596] Change command order in update script.

johngian:


akatsoulas:
How is this going to affect the deployment? r+ for the code

johngian:
``update_info`` calls ``manage.py`` but we activate our ``virtualenv`` on ``setup_dependencies`` meaning that on ``update_info``, ``django`` is not in the path.

PR #1306, opened 2016-01-27
Pin to python 2.7.10 and kombu 2.2.0

johngian:
* We have a hard dependency of celery v2
* The required kombu version needs some deprecated code from python 2.7.10


akatsoulas:
r+

PR #1307, opened 2016-01-27
Run flake8 before installing the packages.

akatsoulas:


johngian:
I suggest we close this now that travis works fine.
@akatsoulas what do you think?

akatsoulas:
I opened this because it makes sense to run flake before pulling the packages so that we can save a few minutes in case flake fails. That said, I don't have strong feelings for this since it's not that important.

johngian:
Ok, let's merge it then. 

* Please rebase.
* Now that we don't have submodules (yeaaay) please remove the ``git submodule update`` lines.

r+wc

PR #1309, opened 2016-01-28
Use pip instead of peep.

akatsoulas:


akatsoulas:
@johngian PR updated and rebased. r? please

johngian:
r+

PR #1310, opened 2016-01-28
Use rabbitmq and celery in docker development environment.

johngian:
* Run celery worker instead of running tasks locally
* Run rabbitmq as a broker for celery tasks


akatsoulas:
r+

PR #1312, opened 2016-02-04
Add cffi and bcrypt to dev requirements.

akatsoulas:
@johngian r?


johngian:
Let's move those requirements to prod.
r+wc

PR #1313, opened 2016-02-04
Small pep8 beautifications.

akatsoulas:
Now with the 99 chars per line a few line breaks can be removed/restructured in favor of code readability. This is a step towards that direction.


akatsoulas:
@johngian r?

johngian:
r+

PR #1318, opened 2016-02-08
Replace 'exact' lookup_type with 'icontains' in Groups API.

akatsoulas:


johngian:
Since ``groups`` and ``skills`` models share some similarities I suggest we do the same to ``skills`` as well.

johngian:
Since we are changing our public api we should update the docs as well.

akatsoulas:
@johngian  pr updated. Thanks!

johngian:
By changing the default lookup type for these fields we are braking the functionality for our current consumers.

I suggest we:
* Leave the default `lookup_type` as it is
* Add an extra `lookup_type` that introduces the new functionality.

akatsoulas:
@johngian pr updated

johngian:
Instead of adding extra fields its better to extra lookup_types. This gives us flexibility to build more on top.

* https://django-filter.readthedocs.org/en/latest/usage.html#the-filter
 * Multiple lookup types

akatsoulas:
@johngian pr updated

mozillians/groups/api/v2.py
```diff
@@ -76,6 +76,14 @@ class Meta:
         model = Group
         fields = ('name', 'functional_area', 'curators', 'curator',
                   'members_can_leave', 'accepting_new_members',)
+        fields = {
+            'name': ['exact', 'icontains'],
+            'functional_area': ['exact'],
```
johngian: Isn't exact the default anyway?

docs/api/apiv2/api-groups.rst
```diff
@@ -22,7 +22,9 @@ Parameters
         *Required* **string** - The application's API key
 
     ``name``
-        *Optional* **string** - Return results matching given name
+        *Optional* **string** - Return results matching the given name
+
+        *Available filters* **icontains** - Return results containing the given name
```
johngian: Please explain how filters are applied in the query.

mozillians/groups/api/v2.py
```diff
@@ -76,6 +76,14 @@ class Meta:
         model = Group
         fields = ('name', 'functional_area', 'curators', 'curator',
```
johngian: This is redefined two lines after.

johngian:
r+

PR #1319, opened 2016-02-09
[fix bug 1241606] Expand settings menu in mobile view.

akatsoulas:
@johngian or @comzeradd r?


comzeradd:
I can't reproduce the bug, so I let @johngian review this. I get all settings without adding the two visibility classes, although I guess it doesn't hurt to add them explicitly.

johngian:
Reproduced it locally and fix seems to be working.

johngian:
Code looks OK. r+

:rocket: :rabbit:

PR #1324, opened 2016-02-17
[Fix bug 1248893] Don't force new curators in legacy groups edit.

johngian:
* Fix initial data in curator form.
* When legacy group, initial should be empty.

### DIFF
--- a/mozillians/groups/forms.py
+++ b/mozillians/groups/forms.py
@@ -24,7 +24,9 @@ class SortForm(forms.Form):
 
 class GroupForm(happyforms.ModelForm):
     curators = forms.ModelMultipleChoiceField(
-        queryset=UserProfile.objects.filter(is_vouched=True).exclude(full_name=''))
+        queryset=UserProfile.objects.filter(is_vouched=True).exclude(full_name=''),
+        required=False
+    )
     invalidation_days = forms.IntegerField(
         widget=forms.NumberInput(attrs={'placeholder': 'days'}),
         min_value=1,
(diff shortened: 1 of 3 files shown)


johngian:
@akatsoulas r?

mozillians/groups/forms.py
```diff
@@ -47,7 +49,11 @@ def clean(self):
                 self._errors['new_member_criteria'] = self.error_class([msg])
                 del cleaned_data['new_member_criteria']
 
-        if not curators:
+        # Check if group is a legacy group without curators.
+        # In this case we should allow curators relation to be empty.
+        allow_no_curators = self.instance.pk and not self.instance.curators.exists()
```
akatsoulas: I would go with `group_has_curators` and remove the `not` both here and a couple of lines below. It's really minor but I think it helps the readability of the code. nice catch btw!

mozillians/groups/views.py
```diff
@@ -397,7 +397,8 @@ def group_add_edit(request, url=None):
 
     form_class = SuperuserGroupForm if is_manager else GroupForm
 
-    curators_ids = [profile.id]
+    curators_ids = [] if group.pk and not group.curators.exists() else [profile.id]
```
akatsoulas: Again here lets reverse the order: [profile.id] if curators.exists() else []

mozillians/groups/tests/test_forms.py
```diff
@@ -41,3 +42,33 @@ def test_no_saved_criteria(self):
         form = GroupForm(data=form_data)
         ok_(form.is_valid())
         eq_(u'', form.cleaned_data['new_member_criteria'])
+
+    def test_legacy_group_curators_validation(self):
```
akatsoulas: nice :sun_with_face:

akatsoulas:
code looks good. Haven't tested this though. r+wc

PR #1329, opened 2016-02-18
Fix initial profiles in curator form.

johngian:
* Group with existing curators must have those as initial
* Group without curators should have request.user as initial
* On ``admins`` or ``managers`` edits lets keep curators intact.

### DIFF
--- a/mozillians/groups/tests/test_views/test_misc.py
+++ b/mozillians/groups/tests/test_views/test_misc.py
@@ -263,3 +263,42 @@ class CreateGroupTests(TestCase):
         ok_(group.visible)
         ok_(group.members_can_leave)
         ok_(not group.functional_area)
+
+    def test_group_add_initial_curators(self):
+        user = UserFactory.create()
+        url = reverse('groups:group_add', prefix='/en-US/')
+
+        with self.login(user) as client:
+            response = client.get(url)
+        eq_(response.status_code, 200)
+
+        initial_curators = response.context['form'].initial['curators']
+        eq_(len(initial_curators), 1)
+        eq_(initial_curators[0], user.userprofile.id)
+
+    def test_group_manager_edit_initial_curators(self):
+        user = UserFactory.create(manager=True)
+        curator = UserFactory.create()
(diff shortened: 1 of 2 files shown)


johngian:
@akatsoulas r?

mozillians/groups/views.py
```diff
@@ -398,7 +398,7 @@ def group_add_edit(request, url=None):
     form_class = SuperuserGroupForm if is_manager else GroupForm
 
     group_has_curators = group.pk and group.curators.exists()
-    curators_ids = [profile.id] if group_has_curators else []
+    curators_ids = [profile.id] if not group_has_curators else []
```
akatsoulas: `not` is not needed here. We need the `profile.id` only and if the group already has curators.

akatsoulas: In order for this to work as described in the comment you need something like:
```python
curators_ids = [group.curators.all().vaulues_list('id', flat=True) if group_has_curators else []
```

mozillians/groups/tests/test_views/test_misc.py
```diff
@@ -263,3 +263,30 @@ def test_group_edit(self):
...
+
+        with self.login(user) as client:
+            response = client.get(url)
+        eq_(response.status_code, 200)
+
+        initial_curators = response.context['form'].initial['curators']
+        eq_(len(initial_curators), 1)
+        eq_(initial_curators[0], user.userprofile.id)
+
+    def test_group_manager_edit_initial_curators(self):
```
akatsoulas: Please enrich this test by adding a second user with an admin role, meaning that the second user should do the edit and should not be in the curators.

mozillians/groups/views.py
```diff
@@ -398,7 +398,7 @@ def group_add_edit(request, url=None):
...
 
     if url:
         curators_ids += group.curators.all().values_list('id', flat=True)
```
akatsoulas: Also remove this line if you go with the suggested change.

akatsoulas:
r+wc :beer: on you :)

PR #1330, opened 2016-02-18
Explicitly define group names in tests.

akatsoulas:
@johngian r?


akatsoulas:
This is rather interesting. The tests were failing because at some point the groups naming schema was reaching number 100. So there were two groups with names `group 99` and `group 100`. The paginator displays the items per page sorted by naming and because `group 100 < group 99` the tests were failing.

johngian:
r+

PR #1335, opened 2016-02-23
[bug 1208367] Raise validation error if skill already exists.

akatsoulas:
@johngian r?

### DIFF
--- a/mozillians/groups/views.py
+++ b/mozillians/groups/views.py
@@ -2,6 +2,7 @@ import json
 
 from collections import defaultdict
 
+from django.http import JsonResponse
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


comzeradd:
Not sure if this is related with this PR, but when I try to create a new skill to add instead of getting just the skill there a `"Create "` prefix.

create

akatsoulas:
This is supposed to happen only if the skill that you are trying to add, does not exist in the db. In your case, if you don't already have `css` in the db, then it's the correct behavior. Otherwise something is not working as it should.

comzeradd:
Code looks good and tested locally. r+ :coffee:

PR #1336, opened 2016-02-23
[bug 1208367] Remove tooltip when adding/editing curators and skills.

akatsoulas:


akatsoulas:
@comzeradd r? 
The problem with the current solution is that the tooltip re-appears when you are editing a skill/curator. This seems to work.

comzeradd:
Code looks and I tested locally. r+

PR #1338, opened 2016-02-24
Don't use the same IDs in group edit template.

akatsoulas:


johngian:
Closing this one after #1339

PR #1341, opened 2016-02-25
Set the max size in multiple curators choicefield.

akatsoulas:
@johngian / @comzeradd r?


johngian:
r+
Tested it locally. ``max-inline-size`` matches the one in the above text area.

PR #1342, opened 2016-02-25
[fix bug 1248893] Force curators for new groups.

akatsoulas:


akatsoulas:
@johngian r?

mozillians/groups/forms.py
```diff
@@ -27,12 +27,13 @@ class GroupForm(happyforms.ModelForm):
...
         min_value=1,
         label='Membership will expire after',
-        required=False
     )
 
     def __init__(self, *args, **kwargs):
         super(GroupForm, self).__init__(*args, **kwargs)
         self.fields['curators'].required = False
+        if not self.instance.pk:
+            self.fields['curators'].required = True
```
johngian: ``self.fields['curators']`` is also required when ``self.instance.pk`` and ``self.instance.groups.exists()``.

johngian:
Please add a new test for that.

mozillians/groups/forms.py
akatsoulas: Sure, this is already handled in the clean method.

akatsoulas:
@johngian pr updated

johngian:
r+

PR #1345, opened 2016-02-25
[bug 1208368] Exclude curators from invalidation.

akatsoulas:


mozillians/groups/tasks.py
```diff
@@ -149,11 +149,14 @@ def invalidate_group_membership():
     groups = Group.objects.filter(invalidation_days__isnull=False)
 
     for group in groups:
+        curator_ids = group.curators.all().values_list('id', flat=True)
+        memberships = (group.groupmembership_set.filter(status=GroupMembership.MEMBER)
+                                                .exclude(curators__id__in=curator_ids))
```
johngian: Query looks wrong here. Group membership doesn't have a ``curators`` field.

mozillians/groups/tasks.py
```diff
@@ -149,11 +149,14 @@ def invalidate_group_membership():
     groups = Group.objects.filter(invalidation_days__isnull=False)
 
     for group in groups:
+        curator_ids = group.curators.all().values_list('id', flat=True)
+        memberships = (group.groupmembership_set.filter(status=GroupMembership.MEMBER)
+                                                .exclude(group__curators__id__in=curator_ids))
```
johngian: This should be something like:
```
memberships=(group.groupmembership_set.filter(status=GroupMembership.MEMBER)                                                .exclude(userprofile__id__in=curator_ids))
```

johngian:
Code looks OK, though I haven't tested it locally.
Please add some tests and make sure that the correct waffle switches are set in prod.

johngian:
r+wc

PR #1348, opened 2016-03-01
Enable invitations in mozillians groups

johngian:


mozillians/groups/models.py
```diff
@@ -330,3 +331,22 @@ def user_can_leave(self, userprofile):
         All users can remove a skill.
         """
         return True
+
+
+class Invite(BaseInvite):
+    group = models.ForeignKey('Group', related_name='invites')
```
akatsoulas: It feels like `group` and `redeemer` should be unique since a user can join a group only once.

mozillians/groups/models.py
```diff
@@ -330,3 +331,22 @@ def user_can_leave(self, userprofile):
...
+    accepted = models.NullBooleanField()
+
+    def __unicode__(self):
+        return 'Group invite {0} for {1}'.format(self.id, self.redeemer)
+
+    def accept(self):
```
akatsoulas: I would suggest to handle this in a `save` method. Feels more natural this way and we can avoid introducing one more method.

mozillians/groups/models.py
```diff
@@ -330,3 +331,22 @@ def user_can_leave(self, userprofile):
...
+        self.group.add_member(self.redeemer, GroupMembership.MEMBER)
+        self.accepted = True
+        self.redeemed = now()
+        self.save()
+
+    def reject(self):
```
akatsoulas: Just delete the invitation. No need to keep these data.

mozillians/groups/urls.py
```diff
@@ -27,6 +27,10 @@
         'views.remove_member', name='remove_member'),
     url('^group/(?P<url>[-\w]+)/confirm/(?P<user_pk>\d+)/$',
         'views.confirm_member', name='confirm_member'),
+    url('^invitation/(?P<invite_pk>\d+)/accept/$',
```
akatsoulas: If you change the `save/accept` and `delete/reject` method, probably you can use only one url with the id. This is going to remove some of the complexity here.

mozillians/phonebook/models.py
```diff
@@ -10,20 +10,28 @@
...
-from mozillians.users.models import UserProfile
 
 
-class Invite(models.Model):
-    inviter = models.ForeignKey(UserProfile, related_name='invites', null=True,
-                                verbose_name=_lazy(u'Inviter'))
-    recipient = models.EmailField(verbose_name=_lazy(u'Recipient'))
-    redeemer = models.OneToOneField(UserProfile, blank=True, null=True,
+class BaseInvite(models.Model):
+    inviter = models.ForeignKey('users.UserProfile',
```
akatsoulas: Probably we should `nullify` this in case the profile of the inviter is deleted.

akatsoulas:
@johngian I had a first pass. Please ping me to have another when you update this PR.

PR #1352, opened 2016-03-03
Don't catch all the exception on celery check.

johngian:
* Raise error for the rest of the exceptions not handled in try/except

### DIFF
--- a/mozillians/users/admin.py
+++ b/mozillians/users/admin.py
@@ -1,4 +1,3 @@
-import sys
 from socket import error as socket_error
 
 from django import forms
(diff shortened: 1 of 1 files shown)


johngian:
@akatsoulas r?

mozillians/users/admin.py
```diff
@@ -474,12 +474,10 @@ def check_celery(self, request):
             investigator.get(timeout=5)
         except investigator.TimeoutError as e:
             messages.error(request, 'Worker timeout: %s' % e)
-        except:
-            messages.error(request, 'An error occured: %s' % sys.exc_info()[0])
         else:
             messages.success(request, 'Celery is OK')
-        finally:
```
akatsoulas: Instead of waiting a 500 error, just leave the general exception and raise explicitly the error.

akatsoulas:
r+wc

PR #1353, opened 2016-03-04
Drop and recreate djcelery table.

akatsoulas:


scripts/update/update.py
```diff
@@ -153,6 +153,8 @@ def update_info(ctx, tag):
         ctx.local("git submodule status")
         ctx.local("which python")
         ctx.local("python manage.py migrate --list")
+        ctx.local("python manage.py migrate djcelery zero")
```
johngian: Please move these lines to the database migration section.

johngian:
r+wc

PR #1354, opened 2016-03-04
[Fix bug 1247221] Implement model backend for group invites.

johngian:
* Expose groups.Invite in /admin
* Add migrations for groups.Invite model


mozillians/groups/admin.py
```diff
@@ -283,6 +284,28 @@ class SkillAdmin(GroupBaseAdmin):
...
+            'inviter': autocomplete.ModelSelect2(url='users:vouched-autocomplete'),
+            'redeemer': autocomplete.ModelSelect2(url='users:vouched-autocomplete'),
+            'group': autocomplete.ModelSelect2(url='groups:group-autocomplete')
+        }
+
+
+class InviteAdmin(ExportMixin, admin.ModelAdmin):
+    search_fields = ['inviter', 'redeemer', 'group']
+    list_display = ['inviter', 'redeemer', 'group']
+    readonly_fields = ['accepted', 'created', 'updated']
```
akatsoulas: I am not against this but do we need the readonly?

mozillians/groups/models.py
```diff
@@ -330,3 +330,23 @@ def user_can_leave(self, userprofile):
...
         return True
+
+
+class Invite(models.Model):
+    inviter = models.ForeignKey('users.UserProfile',
+                                null=True, verbose_name=_lazy(u'Inviter'),
+                                on_delete=models.SET_NULL,
+                                related_name='group_invites')
+    redeemer = models.ForeignKey('users.UserProfile',
+                                 null=True, verbose_name=_lazy(u'Redeemer'),
```
akatsoulas: Are you sure that it's should be `null=True`? The minimum amount of information that we should have is the redeemer and group FKs, and this in only in  the case that the inviter profile gets delete, otherwise we need to keep that too.

mozillians/groups/models.py
```diff
@@ -330,3 +330,23 @@ def user_can_leave(self, userprofile):
...
+                                 related_name='group_invites_redeemed')
```
akatsoulas: Maybe `invites_redeemed`?

mozillians/groups/models.py
```diff
@@ -330,3 +330,23 @@ def user_can_leave(self, userprofile):
         All users can remove a skill.
         """
         return True
+
+
+class Invite(models.Model):
+    inviter = models.ForeignKey('users.UserProfile',
+                                null=True, verbose_name=_lazy(u'Inviter'),
+                                on_delete=models.SET_NULL,
+                                related_name='group_invites')
```
akatsoulas: s/group_invites/invites_sent?

mozillians/groups/models.py
johngian: We also have the same reverse accessor in ``phonebook.Invite``. I would prefer this to be more explicit that it is about groups.

mozillians/groups/models.py
johngian: Same here.

mozillians/groups/admin.py
johngian: I followed the paradigm in ``phonebook.Invite`` I don't have an opinion that. We can make it editable.

akatsoulas:
Sounds good. Let's get rid the `null=True` option

akatsoulas:
r+wc. Please branch out and merge

PR #1359, opened 2016-03-07
[Fix bug 1230887] Add max character limit in search form fields

Ladeia:


johngian:
Code looks OK after the PR update. Thank you for the contribution!
:rocket: 

r+

PR #1360, opened 2016-03-07
[Fix bug 1254127] Make footer sticky to the bottom

comzeradd:


comzeradd:
@akatsoulas r?

johngian:
Please rebase with master.

mozillians/static/mozillians/css/main.less
```diff
@@ -133,8 +140,38 @@
 .html-rtl .right{float:left;}
 .clear{clear:both;}
 
+html, body {
+  height: 100%;
```
johngian: I don't see this getting overridden somehow. Isn't this the default value?

mozillians/static/mozillians/css/main.less
```diff
@@ -133,8 +140,38 @@
...
+}
+
 body {
     #content-wrapper { background: @white; }
+    #outer-wrapper {
+        height: 100%;
```
johngian: Same here.

mozillians/static/mozillians/css/main.less
```diff
@@ -133,8 +140,38 @@
...
+    }
+    #page {
+        min-height: 100%;
```
johngian: Same here.

mozillians/static/mozillians/css/main.less
```diff
@@ -133,8 +140,38 @@
...
+        margin-bottom: -@footerDesktop;
+        @media (max-width: @breakTablet) {
+            margin-bottom: -@footerTablet;
+        }
+        @media (max-width: @breakMobileLandscape) {
+            margin-bottom: -@footerMobile;
+        }
+    }
+    #page:after {
+        content: "";
```
johngian: How does this affect the sticky footer?

mozillians/static/mozillians/css/main.less
```diff
@@ -133,8 +140,38 @@
...
+        display: block;
+        height: @footerDesktop;
+        background-color: @white;
```
johngian: Same here.

mozillians/static/mozillians/css/main.less
```diff
@@ -133,8 +140,38 @@
...
+        min-height: 100%;
+        margin-bottom: -@footerDesktop;
+        @media (max-width: @breakTablet) {
+            margin-bottom: -@footerTablet;
+        }
+        @media (max-width: @breakMobileLandscape) {
+            margin-bottom: -@footerMobile;
+        }
+    }
+    #page:after {
```
johngian: Can't we just add a placeholder element in HTML instead of adding ``:after`` content?

johngian: I am not sure what's better (element vs pseudoelement) please add a comment in the code to explain why we need this.

mozillians/static/mozillians/css/main.less
comzeradd: Not on all browsers.

mozillians/static/mozillians/css/main.less
comzeradd: On div elements the default height is usually auto (depends on the browser), not 100%.

mozillians/static/mozillians/css/main.less
comzeradd: Yes, in general there are two ways of doing this. Either using an empty element with specific height or use an `:after` selector to the existing page wrapper. Both have the same effect. I choose the latter because our wrappers structure permits it and it allows us to keep the html code cleaner.

johngian: Sounds good. Let's add a comment here to explain this for future reference.

mozillians/static/mozillians/css/main.less
```diff
@@ -133,8 +140,38 @@
 .html-rtl .right{float:left;}
 .clear{clear:both;}
 
+html, body {
+  height: 100%;
+}
+
 body {
     #content-wrapper { background: @white; }
+    #outer-wrapper {
```
johngian: Let's concat this one with the same definition above.

comzeradd:
PR rebased to master

mozillians/static/mozillians/css/main.less
```diff
@@ -133,8 +140,38 @@
...
+        @media (max-width: @breakMobileLandscape) {
+            margin-bottom: -@footerMobile;
+        }
+    }
+    // Placeholder for pushing the footer
+    #page:after {
+        content: "";
+        display: block;
+        height: @footerDesktop;
+        background-color: @white;
```
johngian: let's switch the order so ``height`` is right next to the media queries that also change ``height``.

akatsoulas:
The profile view has a line right above the footer.
http://awesomescreenshot.com/0da5rm3lea

akatsoulas:
There is a line in profile_view right above the footer
profile_view

akatsoulas:
The main page when visiting the site seems broken

main_page_logged_out

comzeradd:
@akatsoulas  I fixed the first issue. The problem was that we have the sandstone background to the whole website, instead of just the header where we actually use it.

The second problem is caused due to the fact that the "Get Involved" section is not sticky on its own. In order to stick it to the bottom we should move outside the `#page` wrapper and chage heights and margins specifically for this case (home page and not logged in).

That significantly increases the complexity of the code, but I agree it looks better. I added a second commit for that to see what I mean. I'll squash it if we decide to go this way.

mozillians/jinja2/base.html
```diff
@@ -179,6 +179,8 @@
     {% endblock %}
   </div><!-- close #wrapper -->
 
+  {% block get_involved %}{% endblock %}
+
     {% block site_footer %}
```
johngian: Please fix identation.

johngian:
Logged out page background is broken. (ignore the django error)
selection_024
selection_025

comzeradd:
PR updated. Indentation was off all over the base template.

@johngian I'm having hard time to determine the problem at your last screenshots. Can you post a full page one.

johngian:
selection_026
selection_027

comzeradd:
PR updated with background color set back to sandstone for logged out home page.

ppapadeas:
@comzeradd when logged out, navigating to a profile now displays wrong color for filler. See image
screen shot 2016-04-16 at 13 17 36'

Presumably this happened with the final fix for logged out main page. It should be filling with main color of the page, not hardcoded.

comzeradd:
@ppapadeas good catch. PR updated. I fixed the `#page` background, since the loggedout homepage is the only place we use something else instead of white.

akatsoulas:
@johngian could you please have another pass at this?

johngian:
Please rebase

mozillians/jinja2/base.html
```diff
@@ -57,127 +57,129 @@
...
-                          {% else %}
-                            {{ user.username }}
-                          {% endif %}
+            <ul id="nav-access">
+              <li><a href="#language-switcher">{{ _('Select language') }}</a></li>
+              <li><a href="#content">{{ _('Skip to main content') }}</a></li>
+            </ul>
+            {% block site_header_nav %}
+              <nav role="navigation">
+                  <a class="brand" href="{{ url('phonebook:home') }}">
```
johngian: Since you are fixing the indentation of this file please fix that one as well.

mozillians/jinja2/phonebook/home.html
```diff
@@ -274,6 +254,30 @@
 
 {% endblock content %}
 
+{% block get_involved %}
```
johngian: Since we are using this ``<article>`` only for non-authenticated user's home page, why are you moving it to it's own block that also adds the complexity to check if user is authenticated?

akatsoulas: If this is not needed for the sticky footer, please revert this change.

comzeradd: The "get involved" should be moved between `#page` and `footer`, thus the change. It can't live inside the content block.

comzeradd:
PR rebased. I also fixed indentation issues on base template.

mozillians/jinja2/phonebook/home.html
johngian: sounds good.

akatsoulas:
Tested this locally and looks good. r+ from me

johngian:
On mobile view (360x640) ``#get_involved.content`` is misaligned to the left. It's on logged out home page.

johngian:
r+wc from me

comzeradd:
> On mobile view (360x640) #get_involved.content is misaligned to the left. It's on logged out home page.

I added an explicit `margin-left` to `#get_involved.content` to make sure it's rendered properly on mobile views.

PR #1367, opened 2016-03-09
Setup djcelery.

akatsoulas:


mozillians/settings/base.py
```diff
@@ -469,6 +470,7 @@ def _allowed_hosts():
 DEFAULT_AVATAR_PATH = os.path.join(MEDIA_ROOT, DEFAULT_AVATAR)
 
 # Celery configuration
+djcelery.setup_loader()
```
johngian: I suggest we move this ``wsgi.py``.

akatsoulas: Nah, even django-celery's documentation refers to the `settings.py`. I don't see why we should use this in the wsgi file

johngian:
r+

PR #1371, opened 2016-03-09
Setup django-celery.

akatsoulas:


mozillians/settings/base.py
```diff
@@ -8,8 +8,9 @@
 
 from django.utils.functional import lazy
 
-from urlparse import urljoin
+import djcelery
```
johngian: You are importing djcelery twice.

johngian:
r+wc

mozillians/settings/base.py
johngian: Let's keep the one from ``django-celery`` docs.

PR #1373, opened 2016-03-10
[bug 1254981] Replace cache_page with cache_control.

johngian:
* Instead of using CACHE backend to cache response in the
  view level, add `cache-control` headers.

### DIFF
--- a/mozillians/phonebook/views.py
+++ b/mozillians/phonebook/views.py
@@ -9,7 +9,7 @@ from django.http import HttpResponseRedirect, Http404
 from django.shortcuts import get_object_or_404, render
 from django.utils import timezone
 from django.utils.safestring import mark_safe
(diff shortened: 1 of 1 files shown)


johngian:
@akatsoulas r?

mozillians/phonebook/views.py
```diff
@@ -431,7 +431,7 @@ def betasearch(request):
 
 
 @allow_public
-@cache_page(60 * 60 * 168)  # 1 week.
+@cache_control(must_revalidate=True, max_age=60 * 60 * 168)  # 1 week.
```
akatsoulas: I would suggest to rewrite the `max_age` as `3600 * 24 *7`

akatsoulas: Maybe you can also add `public=True`

akatsoulas:
r+wc :coffee:

PR #1374, opened 2016-03-11
[Fix bug 1255792] Fix humans.txt template path.

johngian:


johngian:
I double checked that the only ``templates`` folder is the one in the top level.

johngian:
@akatsoulas r?

akatsoulas:
r+

PR #1375, opened 2016-03-11
Update all the url patterns in the project.

akatsoulas:


akatsoulas:
I guess we are going to update the urls with the upcoming Django update to the LTS version.

PR #1376, opened 2016-03-15
Do not use the same ID multiple times.

akatsoulas:


akatsoulas:
@johngian r?

johngian:
r+ for removing duplicate IDs.
Just a nit here: Please add distinct IDs (eg: submit-<section>) for completeness.

akatsoulas:
As far as I can tell these IDs are not used anywhere.

PR #1378, opened 2016-03-16
Search user profiles per ID in admin.

akatsoulas:
@johngian r?


johngian:
Why do we need this? You can always get the URL directly.

PR #1384, opened 2016-03-17
Group edit refactor

akatsoulas:

PR #1385, opened 2016-03-17
Add a label in delete profile button.

akatsoulas:


akatsoulas:
@comzeradd r? 
Regarding the font-weight in the label, there is a selector added in PR #1384 which will tackle this one as well.

comzeradd:
r+

PR #1386, opened 2016-03-18
Hide linkedin url and show only identifier in external accounts

Ladeia:


glogiotatidis:
Thanks for the PR @Ladeia. Unfortunately AFAIK not all users in linked-in have an user-set identifier so I don't think this will work.

Ladeia:
hummm, OK.

But what the next step to resolvit?

glogiotatidis:
I don't think that there is something to resolve here. Sadly linked-in urls are like this, there's no username / identifier notion in linkedin

Ladeia:
hummm, 

What do I do in bugzilla bug?

comzeradd:
I think all linkedin users have the option to have a unique username in the form `https://www.linkedin.com/in/<username>`. No? It's not created by default, but it's part of your profile settings on linkedin.

johngian:
What we could do as a workaround is to have a best effort validator that tries to match the username in a linkedin URL. If this is succesfull we should render the username only else we render the whole URL.

Ladeia:
Sorry @johngian I only see your comment now. I can start this modification?

PR #1389, opened 2016-03-21
[Fix bug 1247219] UI fixes for new group settings

comzeradd:
* Minimize space on create group modal
* Use even space on no pending invitations text
* Minimize space on group types descriptions
* Add markdown informative text on terms
* Fix group access criteria toggle

### DIFF
--- a/mozillians/jinja2/groups/includes/group_edit_access.html
+++ b/mozillians/jinja2/groups/includes/group_edit_access.html
@@ -1,5 +1,5 @@
 <div role="tabpanel" class="tab-pane fade" id="access">
-  <form class="edit-group" method="POST" action="{{ url('groups:group_edit', group.url) }}?next=access">
+  <form class="edit-group" id="form-access" method="POST" action="{{ url('groups:group_edit', group.url) }}?next=access">
     {% csrf_token %}
     <div class="panel panel-default">
       <div class="panel-heading">{{ _('Group Type') }}</div>
@@ -25,7 +25,7 @@
           </ul>
         {% endif %}
 
-        <fieldset id="access">
+        <fieldset id="group_type">
           <div class="sizing">
             {{ mozillians_field(criteria_form.accepting_new_members) }}
           </div>
@@ -38,9 +38,7 @@
               Group Curators can invite and/or remove people from the group if needed.
               When in doubt about which group type to choose, default to "Open".
             </p>
-          </div>
 
-          <div class="access-description">
             <h4>Reviewed</h4>
             <p>
               When a group is set to "Reviewed", Mozillians are presented with
@@ -48,9 +46,7 @@
               a pending request notification, prompting them to review and accept or deny the membership.
               Group Curators can invite and/or remove people from the group if needed.
             </p>
-          </div>
 
-          <div class="access-description">
             <h4>Closed</h4>
             <p>
              "Closed" groups provide more tight control over a group.
@@ -63,7 +59,7 @@
 
         <fieldset id="id_new_member_criteria_fieldset">
           <div class="row">
-            <div class="col-md-12">
+            <div class="col-md-8">
               <div class="form-group">
                 {{ mozillians_field(criteria_form.new_member_criteria, required=True) }}
               </div>
@@ -125,9 +121,14 @@
 
           <fieldset id="id_group_terms">
             <div class="row">
-              <div class="col-md-8 col-sm-6">
+              <div class="col-sm-6">
                 <div class="form-group">
                   {{ mozillians_field(terms_form.terms) }}
+                  <span class="under-input-field">
+                    {% trans url='http://daringfireball.net/projects/markdown/syntax' %}
+                      Use <a href="{{ url }}">Markdown</a> for bold, italic, lists and links.
+                    {% endtrans %}
+                  </span>
                 </div>
               </div>
             </div>
@@ -149,7 +150,7 @@
           <fieldset id="id_membership_can_expire">
             <div class="form-group">
               <label class="checkbox">
-                {{ _('Membership can expire:') }}
+                {{ _('Membership can expire') }}
                 <span>
                   <input type="checkbox" class="acknowledge"></input>
                 </span>
--- a/mozillians/jinja2/groups/includes/group_edit_invitation.html
+++ b/mozillians/jinja2/groups/includes/group_edit_invitation.html
@@ -41,7 +41,7 @@
               </div>
             {% endfor %}
           {% else %}
-            <p> This group has no pending invitations.</p>
+            <p class="text-muted">This group has no pending invitations.</p>
           {% endif %}
         </fieldset>
 
--- a/mozillians/static/mozillians/js/group_edit.js
+++ b/mozillians/static/mozillians/js/group_edit.js
@@ -7,7 +7,8 @@ $(function() {
         }
     }
 
-    $id_accepting_new_members = $('#id_accepting_new_members');
+    $group_type_choice = $('#form-access input[type="radio"]');
+    $accepting_new_members = $('input[name=accepting_new_members]:checked');
     $id_new_member_criteria_fieldset = $('#id_new_member_criteria_fieldset');
     $id_invalidation_days = $('#id_group_invalidation_days');
     $id_membership_can_expire = $('#id_membership_can_expire :input');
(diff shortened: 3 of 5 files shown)


comzeradd:
@johngian r?

mozillians/jinja2/groups/includes/group_edit_invitation.html
```diff
@@ -41,7 +41,7 @@
               </div>
             {% endfor %}
           {% else %}
-            <p> This group has no pending invitations.</p>
+            <p class="text-muted">This group has no pending invitations.</p>
```
johngian: ``text-muted`` looks OK. I suggest you also add a CSS rule like the following (with a more appropriate selector):
```
form.edit-group p {
    text-align: center;
}
```

johngian:
New membership terms and membership expiration have different label text format (hint ``:``)

johngian:
I still suggest we consolidate terms/expiration. It doesn't really make any sense to have 2 different panels for these.

@akatsoulas what do you think?

mozillians/jinja2/groups/includes/group_edit_invitation.html
comzeradd: I'm not sure about the alignment. We have no text aligned in center. I think it will look a bit odd.

comzeradd:
I agree. This would require mostly a backend forms consolidation, rather that frontend changes.

mozillians/jinja2/groups/includes/group_edit_invitation.html
johngian: It looks a bit odd to me now. It feels like its a common layout to have centered informative text when a list is empty.

johngian:
I think that the way we handle forms right now in the backend it would be the exact same thing if we post these 2 fields seperately and together.

mozillians/jinja2/groups/includes/group_edit_access.html
```diff
@@ -25,7 +25,7 @@
           </ul>
         {% endif %}
 
-        <fieldset id="access">
+        <fieldset id="access" class="last-fieldset">
```
johngian: Why do we need this class? Also why all that DOM magic when you could select the last fieldset using CSS selectors?

mozillians/static/mozillians/js/group_edit.js
```diff
@@ -18,8 +19,10 @@ $(function() {
     function checkCriteria(modVal) {
         if(modVal === 'by_request') {
             $id_new_member_criteria_fieldset.show();
+            $access.removeClass('last-fieldset');
```
johngian: Same here.

mozillians/static/mozillians/js/group_edit.js
```diff
@@ -18,8 +19,10 @@ $(function() {
...
         } else {
             $id_new_member_criteria_fieldset.hide();
+            $access.addClass('last-fieldset');
```
johngian: Same here.

mozillians/static/mozillians/js/group_edit.js
```diff
@@ -42,10 +45,11 @@ $(function() {
     }
 
     // Hide/show field when moderation field changes
-    $id_accepting_new_members.change(function() {
-        checkCriteria($id_accepting_new_members.val());
+    $('#form-access input[type="radio"]').on('change', function() {
```
johngian: Please use a var for that. Ideally we should use an ``id``.

johngian:
Membership expiration js is also broken.

johngian:
New membership criteria textfield should be under the label like in the rest of the same fields

selection_010

mozillians/static/mozillians/js/group_edit.js
```diff
@@ -42,10 +45,11 @@ $(function() {
...
+        $accepting_new_members = $('input[name=accepting_new_members]:checked');
+        checkCriteria($accepting_new_members.val());
     });
-    $id_membership_can_expire.change(function() {
+    $accepting_new_members.change(function() {
```
johngian: This line here is wrong. You are checking group type to figure out if group membership expires.

akatsoulas: Why is this changed in the first place?

akatsoulas:
:+1: for merging the two sections into one.

mozillians/jinja2/groups/includes/group_edit_access.html
comzeradd: The `:last-child` selector doesn't work here because this is not considered last child, even when the next fieldset is hidden.

mozillians/static/mozillians/js/group_edit.js
comzeradd: Because the previous code was broken. Terms input text is never revealed.

`$accepting_new_members` is `'input[name=accepting_new_members]:checked'`, which just checks which of all three radio buttons is checked.

johngian: Again this is for the group type not for the membership expiration.

mozillians/jinja2/groups/includes/group_edit_invitation.html
comzeradd: We don't use that kind of alignment nowhere else in the site, afaik. Only on alerts.

johngian: Do we have a similar layout somewhere in Mozillians.org so we can follow the same paradigm?

johngian: If not I suggest we go with the centered text.

comzeradd: We have cases where data is pending (like secondary emails or api keys), but not a 100% similar case with this. We just have no centered text in any settings option or text.

johngian: I suggest we go with the centered and let's reevaluate it in the future.

comzeradd:
PR updated. Please test that everything works as expected, especially at the Access tab.

johngian:
Please reduce the space between the 2 checkboxes:
selection_013

johngian:
Also do we need these 2 labels to be bold? (no strong opinion on that)

johngian:
I preferred the previous UI with the choices after the label.
selection_014

johngian:
Also you need to rebase and adapt the code after #1391 fixes.

johngian:
"Membership can expire" is not working.

* Clicked checkbox
* Added days

No data gets saved.

comzeradd:
I fixed the modal misplacement and rebased with master.

I reverted the change with joining terms and expiration. This doesn't make sense to be part of this PR. The forms are separated in the backend and even the tests have to be rewritten.

mozillians/static/mozillians/js/group_edit.js
```diff
@@ -7,9 +7,10 @@ $(function() {
         }
     }
 
-    $id_accepting_new_members = $('#id_accepting_new_members');
+    $group_type_choice = $('#form-access input[type="radio"]');
+    $accepting_new_members = $('input[name=accepting_new_members]:checked');
     $id_new_member_criteria_fieldset = $('#id_new_member_criteria_fieldset');
-    $id_invalidation_days = $('#id_group_invalidation_days');
+    $id_group_invalidation_days = $('#id_group_invalidation_days');
```
johngian: No need to rename the vars here.

mozillians/static/mozillians/js/group_edit.js
```diff
@@ -42,8 +43,9 @@ $(function() {
     }
 
     // Hide/show field when moderation field changes
-    $id_accepting_new_members.change(function() {
-        checkCriteria($id_accepting_new_members.val());
+    $($group_type_choice).on('change', function() {
```
johngian: No need to reselect the selection.

comzeradd:
@johngian PR updated

mozillians/static/mozillians/js/group_edit.js
```diff
@@ -42,8 +43,9 @@ $(function() {
     }
 
     // Hide/show field when moderation field changes
-    $id_accepting_new_members.change(function() {
-        checkCriteria($id_accepting_new_members.val());
+    $group_type_choice.on('change', function() {
+        $accepting_new_members = $('input[name=accepting_new_members]:checked');
```
johngian: Instead of reselecting this here you could use ``this``.

johngian: Instead of copy/pasting the selector here you could reevaluate like:
``$accepting_new_members = $($accepting_new_members.selector)``

johngian:
Code looks OK. r+wc

johngian:
Please test it locally before you merge.

PR #1392, opened 2016-03-21
Multiple fixes for group refactoring

akatsoulas:


mozillians/groups/tests/test_forms.py
```diff
@@ -65,34 +53,113 @@ def test_group_curators_validation(self):
...
 
-    def test_create_new_group(self):
-        # Create a new group without curators
-        curator = UserFactory.create(vouched=True)
-        form_data = {'name': 'test_group',
-                     'accepting_new_members': 'by_request',
-                     'new_member_criteria': 'some criteria',
-                     'curators': [curator.id]}
-        form = GroupForm(data=form_data)
+def test_edit_group_forms(form_class, instance, data, request=None, valid=True):
```
johngian: This function name is kind of confusing since it's not a test. Please rename it to something more relevant eg. ``_edit_group_forms_validation``.

johngian: Also I suggest you move this helper inside ``GroupEditFormTests``.

mozillians/groups/tests/test_forms.py
```diff
@@ -65,34 +53,113 @@ def test_group_curators_validation(self):
...
+
+    def test_edit_invalidation(self):
+        group = GroupFactory.create()
+        data = {'invalidation_days': 5}
+        test_edit_group_forms(forms.GroupInvalidationForm, group, data)
+
+    def test_edit_invitation_without_data(self):
+        group = GroupFactory.create()
+        data = {}
+        test_edit_group_forms(forms.GroupInvalidationForm, group, data)
```
johngian: Here you are testing the ``GroupInvalidationForm`` but the name is about invitations.

johngian:
Since you are changing the check from ``superuser`` to ``manager`` please also change this:
https://github.com/akatsoulas/mozillians/blob/1250876-group-edit-refactor/mozillians/groups/views.py#L518

mozillians/groups/forms.py
```diff
@@ -121,7 +121,7 @@ def clean(self):
         """Custom clean method."""
         super(GroupInviteForm, self).clean()
         user = self.request.user
-        if (self.instance.curators.filter(id=user.userprofile.id).exists() or
+        if (not self.instance.curators.filter(id=user.userprofile.id).exists() and
```
johngian: Let's make this more readable. I suggest we define some vars like ``is_curator`` and ``is_manager`` and check ``if not is_curator and not is_manager).

mozillians/groups/forms.py
```diff
@@ -149,6 +149,17 @@ def __init__(self, *args, **kwargs):
         self.request = kwargs.pop('request', None)
         super(GroupAdminForm, self).__init__(*args, **kwargs)
 
+    def clean(self):
+        """Custom clean method."""
+        super(GroupAdminForm, self).clean()
+        profile = self.request.user.userprofile
+
+        if not profile.is_manager:
+            msg = _(u'You need to be the administrator of this group before '
```
johngian: Let's change this to:
``You need to be the administrator of this group in order to edit this section.``

mozillians/groups/tests/test_forms.py
```diff
@@ -65,34 +53,113 @@ def test_group_curators_validation(self):
...
+    def test_edit_basic_form_with_data(self):
+        group = GroupFactory.create()
+        data = {'name': 'test group',
+                'description': 'sample description',
+                'irc_channel': 'foobar',
+                'website': 'https://example.com',
+                'wiki': 'https://example-wiki.com'}
+        test_edit_group_forms(forms.GroupBasicForm, group, data)
+
+    def test_edit_description_without_data(self):
```
johngian: I think that this should be renamed to:
`` test_edit_basic_form_without_date``.

mozillians/groups/tests/test_forms.py
```diff
@@ -65,34 +53,113 @@ def test_group_curators_validation(self):
...
+        request = RequestFactory().request()
+        request.user = curator
+        data = {'invites': [invitee.userprofile.id]}
+
+        form = test_edit_group_forms(forms.GroupInviteForm, group, data, request)
+        eq_(list(form.instance.invites.all().values_list('id', flat=True)), [invitee.id])
+
+    def test_edit_invitation_without_curator(self):
+        invitee = UserFactory.create()
+        curator = UserFactory.create()
```
johngian: This user is not the curator. Let's rename this to ``user``.

johngian: Also you can always create this directly in:
``request.user = UserFactory.create()``.

mozillians/groups/tests/test_forms.py
```diff
@@ -65,34 +53,113 @@ def test_group_curators_validation(self):
...
+        group = GroupFactory.create()
+        request = RequestFactory().request()
+        request.user = curator
+        data = {'invites': [invitee.userprofile.id]}
+
+        form = test_edit_group_forms(forms.GroupInviteForm, group, data, request, False)
+        eq_(form.errors, {'invites': [u'You need to be the curator of this group before '
+                                      'inviting someone to join.']})
+
+    def test_edit_admin_no_data(self):
```
johngian: Let's keep the same format in all the def names, this should be:
``test_edit_admin_without_data``.

johngian: Also this test is mostly about not having permissions rather than ``without_data``.

johngian:
Do we still need this test:
https://github.com/akatsoulas/mozillians/blob/1250876-group-edit-refactor/mozillians/groups/tests/test_views/test_misc.py#L142

johngian:
r+

akatsoulas:
:beers:

PR #1399, opened 2016-03-23
Exclude request.user from curators/invites autocomplete.

akatsoulas:
@johngian r?


johngian:
We also need to exclude the rest of the curators except of ``request.user``.

akatsoulas:
:+1:

PR #1403, opened 2016-03-24
Remove the readonly field in Invites admin.

akatsoulas:
@comzeradd quick r?

I removed this for two reasons.

1) All the superusers are able to edit this through mozillians.org so it doesn't  really make sense to have this as readonly in admin.

2) It's easier to edit this through admin, especially during dev.


comzeradd:
r+

PR #1405, opened 2016-03-24
Truncate the group name if it is too long in group edit settings.

akatsoulas:


johngian:
code looks ok
r+

johngian:
Please change the commit to mention that you are only changing the name truncation.

johngian:
r+wc

PR #1407, opened 2016-03-28
[fix bug 1259068] Add custom text in group invitation email.

akatsoulas:
@johngian r?

### DIFF
--- a/mozillians/groups/tasks.py
+++ b/mozillians/groups/tasks.py
@@ -167,7 +167,7 @@ def invalidate_group_membership():
 
 
 @task(ignore_result=True)
-def notify_redeemer_invitation(pk):
+def notify_redeemer_invitation(pk, custom_text=''):
 
     from mozillians.groups.models import Invite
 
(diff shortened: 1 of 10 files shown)


mozillians/jinja2/groups/email/invite_email.txt
```diff
@@ -8,6 +8,14 @@ You have been invited by {{ inviter_full_name }} to join
 group "{{ group_name }}" on Mozillians.org.
 {% endtrans %}
 
+{% if custom_text %}
+{% trans %}
+The message from the curator is:
+{% endtrans %}
+
+  {{ custom_text }}
```
johngian: Please fix identation here.

mozillians/groups/tasks.py
```diff
@@ -167,7 +167,7 @@ def invalidate_group_membership():
 
 
 @task(ignore_result=True)
-def notify_redeemer_invitation(pk):
+def notify_redeemer_invitation(pk, custom_text=''):
```
johngian: Instead of ``custom_text=''`` use None and add a check in the template.

johngian:
Please add a test to also use ``custom_text`` when calling tasks.

mozillians/groups/tasks.py
akatsoulas: I chose an empty string as the default value so that we can follow the same paradigm everywhere (eg the default val in the model is an empty string)

akatsoulas:
@johngian pr updated

johngian:
Tested it locally. r+

PR #1409, opened 2016-03-28
[Fix bug 1259453] Remove extra space from links

comzeradd:


comzeradd:
@johngian r?

mozillians/jinja2/groups/includes/group_edit_invitation.html
```diff
@@ -19,11 +19,11 @@
                 <div class="col-sm-8">
 
                   <div class="avatar">
+                    <img class="profile-photo"
+                        src="{{ redeemer.get_photo_url('70x70') }}"
+                        alt="{{ _('Profile Photo') }}"
+                        title={{ redeemer.display_name }}>
                     <a href="{{ url('phonebook:profile_view', redeemer.user.username) }}">
-                      <img class="profile-photo"
-                          src="{{ redeemer.get_photo_url('70x70') }}"
```
johngian: Please indent. (eg. ``src`` should be in the same column with ``class``).

johngian:
r+wc

johngian:
FYI you could have used this instead which is the jinja2 way to fix that kind of issues.
http://jinja.pocoo.org/docs/dev/templates/#whitespace-control

PR #1410, opened 2016-03-29
[bug 1247219] Capitalize first letters in group edit submit buttons.

akatsoulas:


johngian:
Let's fix this one as well:
https://github.com/akatsoulas/mozillians/blob/edit-groups-fix-typo/mozillians/jinja2/groups/includes/group_edit_access.html#L174

comzeradd:
If we want to capitalize first letters on submit buttons I suggest we do it everywhere. Besides the one @johngian already pointed there is this: https://github.com/akatsoulas/mozillians/blob/edit-groups-fix-typo/mozillians/jinja2/groups/includes/group_edit_admin.html#L36

akatsoulas:
@johngian pr updated

johngian:
r+

PR #1417, opened 2016-03-29
[fix bug 1258324] Allow all mozillians to accept/reject invitations.

akatsoulas:
@johngian r?

### DIFF
--- a/mozillians/groups/views.py
+++ b/mozillians/groups/views.py
@@ -166,14 +166,14 @@ def show(request, url, alias_model, template):
 
             # Curators can delete their group if there are no other members.
             show_delete_group_button = is_curator and group.members.all().count() == 1
-            invitation = get_object_or_none(Invite, redeemer=profile, group=group, accepted=False)
-            data.update(invitation=invitation)
 
         else:
             # only show full members, or this user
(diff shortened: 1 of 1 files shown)


mozillians/groups/views.py
```diff
@@ -166,14 +166,14 @@ def show(request, url, alias_model, template):
...
             show_delete_group_button = is_curator and group.members.all().count() == 1
-            invitation = get_object_or_none(Invite, redeemer=profile, group=group, accepted=False)
-            data.update(invitation=invitation)
 
         else:
             # only show full members, or this user
             memberships = group.groupmembership_set.filter(
                 Q(status=GroupMembership.MEMBER) | Q(userprofile=profile))
 
+        invitation = get_object_or_none(Invite, redeemer=profile, group=group, accepted=False)
```
johngian: I think that you should check here if the group that we are trying to render is a ``Group`` instance and not a ``Skill``.

akatsoulas: this is inside the if clause that checks for a group instance

johngian: :+1:

johngian:
r+

PR #1418, opened 2016-03-29
[Bug 1247219] Fix Group settings UI elements

comzeradd:
* Fix Group title overflow text
* Limit select boxes width on mobile
* Proper arrangement of checkbox labels

### DIFF
--- a/mozillians/static/mozillians/css/main.less
+++ b/mozillians/static/mozillians/css/main.less
@@ -299,6 +299,13 @@ body {
     .select2-container--default .select2-selection--multiple .select2-selection__clear {
         margin-right: 0;
     }
+
+    @media (max-width: @breakMobileLandscape) {
+        .select2-selection--multiple,
+        .select2-dropdown.select2-dropdown--below {
+            max-width: 230px !important;
+        }
+    }
 }
 
 #outer-wrapper {
@@ -2384,6 +2391,12 @@ form.add-group {
             padding-left: 0px;
         }
     }
+    #id_group_email_text,
+    #id_group_terms,
+    #id_group_invalidation_days {
+        margin-top: 0px;
+        padding-top: 0px;
+    }
     .text-muted {
         margin: 20px 0;
         text-align: center;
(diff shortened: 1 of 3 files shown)


comzeradd:
@johngian r?

johngian:
Tested it locally

* Checkboxes are working fine
* Curators autocomplete field looks OK on complete
* Overflow text looks OK

johngian:
Dropdown on curators auto-completion overflows field above.

johngian:
Delete/resend invitation buttons are misaligned in mobile view (eg. 320x480)

johngian:
Invites auto-completion field has the same issue that curators had.

johngian:
I suggest we reduce the height between invites - checkbox and checkbox - textarea. @comzeradd what do you think?

selection_020

comzeradd:
> Delete/resend invitation buttons are misaligned in mobile view (eg. 320x480)

Can you share a screenshot? If you mean that these are displayed in a new line that's expected. They stack under the name.

johngian:
Is this expected? It's fine by  me if yes, but i thought this was misaligned:

selection_021

comzeradd:
Yes that's expected. In most cases the name would be long enough anyway. It only looks kind of strange on very small names.

PR updated. I fixed the dropdown overflows. Unfortunately that needs an `!important` statement because the select2 js adds inline style. I also minimize the space between invitation email checkbox and textarea.

johngian:
selection_022

Do you think that we need to add some padding here (check username)? The same applies for curators. (resolution 320x480)

johngian:
Verified locally the rest of the changes and look OK.

comzeradd:
Yes, you're right. I missed the help-block. I pushed a minor change to have the same width as the select input.

mozillians/static/mozillians/css/main.less
```diff
@@ -2384,6 +2391,13 @@ form.add-group {
             padding-left: 0px;
         }
     }
+    #id_group_email_text {
+        margin-top: 0px;
+        padding-top: 0px;
+    }
+    .help-block {
+        max-width: 230px;
```
akatsoulas: This now forces the help text to break into multiple lines even when is not needed (eg full screen)

comzeradd:
@akatsoulas PR updated

akatsoulas:
Let's apply what you did in the custom email text in the other checkboxes too. r+wc

PR #1420, opened 2016-03-30
[fix bug 1260429] Fix the url for displaying skills.

akatsoulas:


akatsoulas:
@johngian r?

johngian:
Please add a test for ``get_absolute_url``.

mozillians/groups/models.py
```diff
@@ -25,6 +26,14 @@ class Meta:
         abstract = True
         ordering = ['name']
 
+    def get_absolute_url(self):
+        if isinstance(self, Group):
+            return absolutify(reverse('groups:show_group', args=[self.url]))
+        elif isinstance(self, Skill):
+            return absolutify(reverse('groups:show_skill', args=[self.url]))
+        else:
```
johngian: Please remove else case. In case ``get_absolute_url`` fails it should through its own exception.

johngian:
r+wc

mozillians/groups/models.py
johngian: Instead of overriding this with ``if/elif`` lets be more explicit and override ``get_absolute_url`` in both ``Group`` and ``Skill`` classes.

johngian: This way you can also avoid writing extra tests.

akatsoulas: Nah, I prefer this to be DRYer. There is no need to duplicate code since all the child classes can inherit this. :+1: for the tests

mozillians/groups/models.py
```diff
@@ -25,6 +25,11 @@ class Meta:
         abstract = True
         ordering = ['name']
 
+    def get_absolute_url(self):
+        cls_name = self.__class__.__name__
+        return absolutify(reverse('groups:show_{0}'.format(cls_name.lower()),
```
johngian: Please split these to 2 lines to make it more readable.

johngian:
Tests are failing locally.

akatsoulas:
Yap, that's why I am not merging this yet. Trying to figure this out

PR #1422, opened 2016-03-30
[fix bug 1260675] Split the invitation form in group edit.

akatsoulas:


akatsoulas:
@johngian r?

mozillians/groups/forms.py
```diff
@@ -135,16 +135,20 @@ def clean(self):
 
     def save(self, *args, **kwargs):
         """Custom save method to add data to the through model."""
-        custom_text = self.cleaned_data['invite_email_text']
-        self.instance.invite_email_text = custom_text
-        self.instance.save()
+
+        saved_instance = None
```
johngian: I would prefer something more specific like:

```
invite_email_text = ""
if self.instance.pk:
    invite_email_text = self.instance.invite_email_text
```

I would also reduce the scope of this to line 151.

johngian: Also even if instance is not saved, ``self.instance.invite_email_text`` will be ``""`` because of default. I assume that you could avoid all the checks and directly use the attribute.

mozillians/groups/tests/test_forms.py
```diff
@@ -176,16 +179,10 @@ def test_edit_admin(self):
...
-        invitee = UserFactory.create()
         curator = UserFactory.create()
         group = GroupFactory.create()
         group.curators.add(curator.userprofile)
         request = RequestFactory().request()
         request.user = curator
-        data = {'invites': [invitee.userprofile.id],
-                'invite_email_text': u'Custom message in the email.'}
-
-        with patch('mozillians.groups.forms.notify_redeemer_invitation.delay') as mocked_task:
```
johngian: Why are you removing this line. It would be helpful to also test the mocked task call.

akatsoulas: This check is not relocated to the `test_edit_invitation` test. In this test all we do is to test that the form and its data are valid. This form does not trigger any tasks/emails.

mozillians/groups/forms.py
akatsoulas: This was here for the wrong reasons. This is now fixed.

akatsoulas:
@johngian pr updated

johngian:
r+

PR #1425, opened 2016-04-04
Fix docker build issues.

johngian:
* `apt-get update` gets cached and in case it fails the whole build
  fails
* It's best practice to run all `apt` related commands to a single RUN


johngian:
@akatsoulas r?

akatsoulas:
r+

PR #1427, opened 2016-04-05
[fix bug 1262079] Fix unicode error in group tasks.

akatsoulas:
@johngian r?


johngian:
Let's add a test to also check the regression from the traceback.

johngian:
r+wc

PR #1435, opened 2016-04-13
Add some more files in gitignore.

akatsoulas:


.gitignore
```diff
@@ -12,7 +12,9 @@ docs/_gh-pages
 docs/_build
 locale*
 /venv/
+/virtualenv/
 /static/
-/media/uploads/
+/media/*
```
johngian: We don't really want to stop tracking everything under ``/media/``. eg
```
$ git ls-files media/
media/img/default_avatar.png
media/robots-prod.txt
media/robots.txt
```

.gitignore
```diff
@@ -12,7 +12,9 @@ docs/_gh-pages
...
 /elasticsearch*/
 mozillians/settings/local.py
+/vendor/
```
johngian: Instead of adding ``/vendor/`` here I suggest we open a bug to clean up the server files. The same applies for the submodule we had under static: ``mozillians/static/mozillians/css/less-prefixer/``

akatsoulas:
The bug sounds good to me.

PR #1439, opened 2016-04-16
Fix the ordering of 'number of vouches' in users admin.

akatsoulas:


akatsoulas:
@johngian r?

johngian:
Code looks OK. I also tested it locally and it works fine.

johngian:
Please open a bug for this issue so we can track and update commit message.

johngian:
r+wc

PR #1441, opened 2016-04-29
Use HTTPS scheme for contribute URL

stephendonner:
This hardcoded HTTP URL and another (the privacy_policy_url in https://github.com/mozilla/mozillians/blob/1ebc088c297a245bd8b156b0e4a3617f38c99c8b/mozillians/jinja2/phonebook/about.html#L37-L38) are coming up in another bug as a small issue (can catch you up, offline).  Should be a safe change.

r? @akatsoulas @johngian


claudijd:
👍

akatsoulas:
Thanks @stephendonner r+

PR #1442, opened 2016-04-29
Use HTTPS scheme for privacy URL

stephendonner:
Should be safe to default to HTTPS (and avoids an extra HTTP redirect).


akatsoulas:
r+

PR #1446, opened 2016-05-05
Add missing migrations.

akatsoulas:
@johngian can you double check and r this?


johngian:
I verified it locally by doing a clean docker build. It looks like we forgot to ``makemigrations`` after a ``pytz`` upgrade.

johngian:
r+

PR #1451, opened 2016-05-12
[Bug 1254127] Fix get involved margins on mobile

comzeradd:


comzeradd:
@johngian r?

johngian:
Using the inspector it looks like ``section#about`` has ``margin-left: 20px``. Why did you decided to explicitly add ``margin-left: 10px`` ?

comzeradd:
I did so to avoid complicating the code with multiple height options for making it sticky, since the margin will affect `get_involved` section wrapping.

To make it align with `#about` above I just did a small adjustment in order to have the same margin on both.

mozillians/static/mozillians/css/main.less
```diff
@@ -1200,6 +1200,7 @@ body#home {
                 @media (max-width: @breakMobileLandscape) {
                     width: auto;
                     padding: 40px 0 20px;
+                    margin: 0 10px;
```
akatsoulas: Why don't you just set the left margin here to `20px` to match the above?

comzeradd: That would cause text to overlap and would require more changes. Besides 20px seems too much anyway, since these elements have an inherited padding from the container..

akatsoulas: I am not sure that I follow. Can you please collaborate on which text are you referring to? Testing this with a left margin of 20px (in the media query) works OK for me. Could you please provide a screenshot? Regarding the 20px value itself, I just recommended that because this is the value already used in the `about` section.

comzeradd: I attach a screenshot of how it looks on 320 with 20px margins. Regardless of that, as I already mentioned, there is no need for having so much space on both sides.

screen shot 2016-05-19 at 15 33 13

johngian:
@comzeradd 
Although I don't really agree with the approach of changing the section above just because the wrapping below doesn't get fixed automatically, let's merge it since its a super minor issue.

johngian:
r+

comzeradd:
Thanks. The major reason for changing the margin is because it's too much. The benefit of less complexity on sticking the `get_involved` section is just a side bonus.

PR #1452, opened 2016-05-13
Remove 'http' entries from CSP.

akatsoulas:


johngian:
Doubled checked there are no references to those entries. The issue is going to be in our local envs that will not be able to load assets that are fetched depending in the site url (eg starting with "//")

akatsoulas:
We could add these entries in the `local.py-dist`. Thoughts?

johngian:
sounds good

akatsoulas:
@johngian pr updated

johngian:
``
  File "/code/mozillians/settings/local.py", line 90, in <module>
    base.CSP_DEFAULT_SRC += ('http://*.mapbox.com',)
NameError: name 'base' is not defined
``

akatsoulas:
Ermmm, my bad, pr updated.

mozillians/settings/local.py-dist
```diff
@@ -53,4 +53,17 @@ CACHES = {
         'LOCATION': 'unique-snowflake'
     }
 }
-BROWSERID_AUDIENCES = None
\ No newline at end of file
+BROWSERID_AUDIENCES = None
+base.CSP_DEFAULT_SRC += ('http://*.mapbox.com',)
```
johngian: You don't really override the values here. Try something like:
```

CSR_DEFAULT_SRC = base.CSP_DEFAULT_SRC + ('http://*.mapbox.com',)
CSP_FONT_SRC = base.CSP_FONT_SRC +('http://*.mozilla.net',
                                   'http://*.mozilla.org',
                                   'http://mozorg.cdn.mozilla.net',)
CSP_IMG_SRC = base.CSP_IMG_SRC + ('http://*.mozilla.net',
                                  'http://*.mozilla.org',
                                  'http://*mapbox.com',)
CSP_SCRIPT_SCR = base.CSP_SCRIPT_SRC + ('http://*.mozilla.net',
                                        'http://*.mozilla.org',
                                        'http://*mapbox.com',)
CSP_STYLE_SRC = base.CSP_STYLE_SRC + ('http://*.mozilla.net',
                                      'http://*.mozilla.org',
                                      'http://*mapbox.com',)
```

mozillians/settings/local.py-docker-dist
```diff
@@ -85,4 +86,18 @@ CACHES = {
     }
 }
 
-BROWSERID_AUDIENCES = None
\ No newline at end of file
+BROWSERID_AUDIENCES = None
+
+base.CSP_DEFAULT_SRC += ('http://*.mapbox.com',)
```
johngian: Same here.

johngian:
r+

glogiotatidis:
Just curious: How do you handle `//example.com/foo.js` links during development where you don't have `https` and therefore the example translates to `http://example.com/foo.js`

johngian:
This should be addressed after overriding CSP related configuration entries in ``settings/local.py-{dist, docker-dist}``.

glogiotatidis:
ic, that's tricky. Thanks for the answer!

akatsoulas:
@glogiotatidis can you please elaborate more about the tricky part? Do you have any ideas for a cleaner solution?

glogiotatidis:
I consider that tricky because you'll not face CSP errors during development and the possibilities to land code that doesn't work on stage is higher. Unless you actually maintain the `http` version of CSP rules on your local file.

I prefer to keep `http` the CSP rules in place, since we run everything under `https` on production I don't see how they can be a problem.

akatsoulas:
mozillians.org was allowing `http` on prod  thus this change. Thanks for the explanation! :beer:

johngian:
FYI this was implemented after some security suggestions not to allow CSP content using ``http``. That's why the explicit CSP rule.

PR #1455, opened 2016-05-20
Remove 'git submodule' entries from chief script.

akatsoulas:


johngian:
We are not currently using any submodules. Chief script looks OK. r+

PR #1458, opened 2016-05-30
[Fix bug 1276615] Define `functional_areas` before rendering template

johngian:


akatsoulas:
Please keep the commit message below 80 chars. r+wc :+1:

PR #1459, opened 2016-06-02
Replace peep with pip hash checking when building docker.

johngian:


johngian:
@akatsoulas r?

Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM python:2.7.10
 ADD . /code
 WORKDIR /code
-RUN apt-get update && apt-get install -y mysql-client node-less
-RUN ./peep.sh install -r requirements/dev.txt
\ No newline at end of file
+RUN apt-get update && apt-get install -y mysql-client node-less &&\
+    pip install --upgrade pip
```
akatsoulas: Maybe we should use something here like `pip install -U pip==8` (not a blocker though)

akatsoulas:
r+wc

PR #1460, opened 2016-06-06
Implement SRI mixin for django-compressor

johngian:
* Implement custom css/js processor with sri support.
* Override compressor templates to support sri.
* Add settings to use custom js/css compressors

### DIFF
--- a/mozillians/common/compress.py
+++ b/mozillians/common/compress.py
@@ -1,0 +1,53 @@
+from __future__ import with_statement, unicode_literals
+
+from django.core.files.base import ContentFile
+from django.utils.safestring import mark_safe
+
+import base64
+
+from compressor.css import CssCompressor
+from compressor.js import JsCompressor
+from hashlib import sha384
+
+
+def _compute_sri_value(content):
+    sri_hash = sha384(content.encode('utf-8')).digest()
+    sri_value = base64.b64encode(sri_hash)
+    return 'sha384-{}'.format(sri_value)
+
+
+class SRIMixin(object):
+
+    def output_file(self, mode, content, forced=False, basename=None):
+        """
+        The output method that saves the content to a file and renders
+        the appropriate template with the file's URL.
+        """
+        new_filepath = self.get_filepath(content, basename=basename)
+        if not self.storage.exists(new_filepath) or forced:
+            self.storage.save(new_filepath, ContentFile(content.encode(self.charset)))
+        url = mark_safe(self.storage.url(new_filepath))
+        context = {
+            'url': url,
+            'sri': _compute_sri_value(content)
+        }
+        return self.render_output(mode, context)
+
+    def output_inline(self, mode, content, forced=False, basename=None):
+        """
+        The output method that directly returns the content for inline
+        display.
+        """
+        context = {
+            'content': content,
+            'sri': _compute_sri_value(content)
+        }
+        return self.render_output(mode, context)
+
+
+class SRIJsCompressor(SRIMixin, JsCompressor):
+    pass
+
+
+class SRICssCompressor(SRIMixin, CssCompressor):
+    pass
--- a/mozillians/templates/compressor/css_file.html
+++ b/mozillians/templates/compressor/css_file.html
@@ -1,0 +1,1 @@
+<link rel="stylesheet" href="{{ compressed.url }}" type="text/css"{% if compressed.media %} media="{{ compressed.media }}"{% endif %} {% if compressed.sri %}integrity="{{ compressed.sri }}" crossorigin="anonymous"{% endif %}/>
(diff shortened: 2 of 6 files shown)


akatsoulas:
nice :+1:

johngian:
This is currently a WIP. I've tested it locally and it looks like it's working (both offline and online compression). Some thoughts:

* Integrity doesn't currently work when ``DEBUG = True``. At first it looked like it was needed and spent some time to fix it too. In the end it got really complicated and after digging more into ``django-compressor`` code many things don't really work on development mode.

* The hashing algorithm is currently hardcoded. ``SRI`` supports many different algorithms but I choose the one that is proposed in MDN (sha384). Do you think that it makes sense to make it more generic and add some config to choose the hashing algorithm?

My initial idea was to make something that it's stable enough for mozillians.org and then make it more generic to tackle https://github.com/django-compressor/django-compressor/issues/661. I still don't know if I am on the right track even though it looks like it's working fine.

@akatsoulas thoughts?

rfleschenberg:
This computes the hash on each request, right? Could this become a performance problem?

johngian:
@rfleschenberg in our setup we are using offline compression so this gets computed only once

mozillians/common/compress.py
```diff
@@ -0,0 +1,53 @@
+from __future__ import with_statement, unicode_literals
+
+import base64
+
+from compressor.css import CssCompressor
```
glogiotatidis: Third party libraries after django imports or not?

mozillians/common/compress.py
```diff
@@ -0,0 +1,53 @@
...
+from hashlib import sha384
+
+from django.core.files.base import ContentFile
+from django.utils.safestring import mark_safe
+
+
+def _compute_sri_value(content):
+    sri_hash = sha384(content.encode('utf-8')).digest()
+    sri_value = base64.b64encode(sri_hash)
+    return "sha384-{}".format(sri_value)
```
glogiotatidis: nit: single quotes instead of douple

mozillians/common/compress.py
```diff
@@ -0,0 +1,53 @@
...
+from compressor.css import CssCompressor
+from compressor.js import JsCompressor
+from hashlib import sha384
+
+from django.core.files.base import ContentFile
+from django.utils.safestring import mark_safe
+
+
+def _compute_sri_value(content):
+    sri_hash = sha384(content.encode('utf-8')).digest()
```
glogiotatidis: Convert this to python3 compat code if you're PR upstream.

mozillians/common/compress.py
```diff
@@ -0,0 +1,53 @@
...
+
+from compressor.css import CssCompressor
+from compressor.js import JsCompressor
+from hashlib import sha384
+
+from django.core.files.base import ContentFile
+from django.utils.safestring import mark_safe
+
+
+def _compute_sri_value(content):
```
glogiotatidis: Since we're using offline compression this doesn't hurt but you could wrap with with Django's `lru_cache` implementation to provide a solution when doing online compression

https://github.com/django/django/blob/master/django/utils/lru_cache.py

mozillians/common/compress.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        The output method that saves the content to a file and renders
+        the appropriate template with the file's URL.
+        """
+        new_filepath = self.get_filepath(content, basename=basename)
+        if not self.storage.exists(new_filepath) or forced:
+            self.storage.save(new_filepath, ContentFile(content.encode(self.charset)))
+        url = mark_safe(self.storage.url(new_filepath))
+        context = {
+            "url": url,
```
glogiotatidis: nit: quotes here and in other places

mozillians/templates/compressor/css_file.html
```diff
@@ -0,0 +1 @@
+<link rel="stylesheet" href="{{ compressed.url }}" type="text/css"{% if compressed.media %} media="{{ compressed.media }}"{% endif %} integrity="{{ compressed.sri }}" crossorigin="anonymous"/>
```
glogiotatidis: Maybe you wrap `integrity="{{ compressed.sri }}"` with `{% if compressed.sri %}` since I understand that the changes override compressor templates no matter the compressor engine used.

So for example if in production you revert COMPRESS_CSS_COMPRESSOR back to django_compressor's default you don't want this to break things.

glogiotatidis:
Great work @johngian. Eager to see some of our websites using SRI. 🍻

mozillians/common/compress.py
glogiotatidis: Nope, this is fine as is.

PR #1461, opened 2016-06-08
[Fix bug 1276614] Use same height for header

comzeradd:


comzeradd:
@johngian r?

johngian:
Code looks OK. Tested it locally for both desktop and mobile and looks fine.
r+

PR #1462, opened 2016-06-09
Always use HTTPS when declaring css fonts.

johngian:


johngian:
This came up while reviewing #1461

It looks like this fixes the ``CORS`` errors in local development. Also I don't see what's the point of serving non-HTTPS assets.

akatsoulas:
r+

PR #1467, opened 2016-06-29
Refactor basket tasks to take `newsletters` as input

johngian:
* Extend tasks to take `newsletters` as arg
* Add new `settings` entry for NDA basket
* When user changes email re-subscribe to all newsletters not only vouched

### DIFF
--- a/mozillians/users/admin.py
+++ b/mozillians/users/admin.py
@@ -46,7 +46,7 @@ def subscribe_to_basket_action():
     def subscribe_to_basket(modeladmin, request, queryset):
         """Subscribe to Basket or update details of already subscribed."""
         ts = [(mozillians.users.tasks.update_basket_task
-               .subtask(args=[userprofile.id]))
+               .subtask(args=[userprofile.id, [settings.BASKET_VOUCHED_NEWSLETTER]]))
               for userprofile in queryset]
         TaskSet(ts).apply_async()
         messages.success(request, 'Basket update started.')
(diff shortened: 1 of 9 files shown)


mozillians/users/admin.py
```diff
@@ -46,7 +46,7 @@ def subscribe_to_basket_action():
     def subscribe_to_basket(modeladmin, request, queryset):
         """Subscribe to Basket or update details of already subscribed."""
         ts = [(mozillians.users.tasks.update_basket_task
-               .subtask(args=[userprofile.id]))
+               .subtask(args=[userprofile.id, [settings.BASKET_VOUCHED_NEWSLETTER]]))
```
akatsoulas: Shouldn't this action subscribe all the users in the queryset in both NDA (if and only if the user is a member of the nda group) and the VOUCHED newsletters?

mozillians/users/admin.py
```diff
@@ -60,7 +60,8 @@ def unsubscribe_from_basket_action():
     def unsubscribe_from_basket(modeladmin, request, queryset):
         """Unsubscribe from Basket."""
         ts = [(mozillians.users.tasks.unsubscribe_from_basket_task
-               .subtask(args=[userprofile.user.email, userprofile.basket_token]))
+               .subtask(args=[userprofile.user.email, userprofile.basket_token,
```
akatsoulas: Similar here

mozillians/users/admin.py
johngian: This is temp code just to cleanup the PR. Any admin related functionality should be tackled in another bug PR.

johngian: What i am doing here is replicating the functionality in ``master`` using the generic basket tasks.

akatsoulas: If this is the case, I would suggest to refactor the `update_basket_task`  so that the newsletters arg is an empty/optional argument like:

```python
def update_basket_task(instance_id, newsletters=[]):
    if not newsletters:
        return
```
This way you won't have to deal with this code snippet at all and hopefully catch a few errors in the future.

akatsoulas:
r+wc, nice work :airplane:

mozillians/users/admin.py
akatsoulas: Or raise an exception here. If you do please add a test to demonstrate this too :smile:

PR #1469, opened 2016-07-04
[Fix bug 1284254] Remove vouch check on basket subscription.

johngian:


akatsoulas:
Also add a test without newsletter and verify that the `basket.subscribe` has no calls. r+wc :+1:

PR #1470, opened 2016-07-04
[fix bug 1282452] Update basket for nda subscription.

akatsoulas:


akatsoulas:
@johngian r?

mozillians/groups/models.py
```diff
@@ -274,6 +278,10 @@ def add_member(self, userprofile, status=GroupMembership.MEMBER):
                 membership.save()
                 if membership.status in [GroupMembership.PENDING, GroupMembership.MEMBER]:
                     email_membership_change.delay(self.pk, userprofile.user.pk, old_status, status)
+                # Since there is no demotion, we can check if the new status is MEMBER and
+                # subscribe the user to the NDA newsletter if the group is NDA
+                if status == GroupMembership.MEMBER and self.name == settings.NDA_GROUP:
```
johngian: nit: Let's switch the order of the checks here. It's much more often for ``self.name == settings.NDA_GROUP`` rather than ``status == GroupMembership.MEMBER``. This way we use the python ``and`` short-circuiting functionality and do less assertions.

mozillians/groups/tests/test_models.py
```diff
@@ -236,32 +236,62 @@ def test_deleted_curator_sets_null(self):
...
+    def test_remove_member(self, unsubscribe_basket_mock):
         user = UserFactory.create()
         group = GroupFactory.create()
         GroupMembership.objects.create(userprofile=user.userprofile, group=group,
                                        status=GroupMembership.MEMBER)
         ok_(group.has_member(user.userprofile))
 
         group.remove_member(user.userprofile)
+        ok_(not unsubscribe_basket_mock.called)
         ok_(not GroupMembership.objects.filter(userprofile=user.userprofile,
```
johngian: Please add a comment to explain here why mock is not called or explicitly add a group name different of ``nda``. Or both.

mozillians/groups/tests/test_models.py
```diff
@@ -236,32 +236,62 @@ def test_deleted_curator_sets_null(self):
...
         ok_(not group.has_member(user.userprofile))
 
-    def test_add_member(self):
+    @patch('mozillians.users.tasks.unsubscribe_from_basket_task')
+    @override_settings(BASKET_VOUCHED_NEWSLETTER='nda')
+    def test_remove_member_nda(self, unsubscribe_basket_mock):
+        user = UserFactory.create(userprofile__email='user@example.com',
+                                  userprofile__basket_token='basket_token')
+        group = GroupFactory.create(name='nda')
+        GroupMembership.objects.create(userprofile=user.userprofile, group=group,
```
johngian: Why do you both create``GroupMembership`` insttance and call ``group.add_members``?

johngian:
Let's also add a test for cases that user was member of group ``NDA`` but now state is pending.

johngian:
r+wc

PR #1482, opened 2016-07-13
[Fix bug 1286559] Fix update basket subscriptions task.

johngian:
* Check first for email changes
* Subscribe to required newsletters
* Update token if no token stored for mozillian yet

### DIFF
--- a/mozillians/users/tasks.py
+++ b/mozillians/users/tasks.py
@@ -91,31 +91,7 @@ def update_basket_task(instance_id, newsletters=[]):
     email = instance.user.email
     token = instance.basket_token
 
-    if not token:
-        # no token yet, they're probably not subscribed, so subscribe them.
-        # pass sync='Y' so we wait for it to complete and get back the token.
-        try:
-            retval = basket.subscribe(
-                email,
-                newsletters,
-                sync='Y',
-                trigger_welcome='N'
-            )
-        except (requests.exceptions.RequestException,
-                basket.BasketException) as exception:
-            try:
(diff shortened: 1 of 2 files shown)


mozillians/users/tasks.py
```diff
@@ -163,6 +140,28 @@ def update_basket_task(instance_id, newsletters=[]):
             # via the post-save signal, which would be pointless.
             UserProfile.objects.filter(pk=instance.pk).update(basket_token=token)
 
+    try:
+        retval = basket.subscribe(
```
johngian: Check first if ``newsletters`` already exist in ``lookup_user`` results.

mozillians/users/tasks.py
```diff
@@ -163,6 +140,28 @@ def update_basket_task(instance_id, newsletters=[]):
             # via the post-save signal, which would be pointless.
             UserProfile.objects.filter(pk=instance.pk).update(basket_token=token)
 
+    try:
```
akatsoulas: This `try/catch` clause will try to subscribe the user even if there is already a token. Shouldn't we try a subscription only in the case that a user does not have a token?

johngian: No because the user might already have a token from an "X" newsletter subscription and we also want to subscribe him to newsletter "Y". Tokens are per user not per subscription.

johngian: What you are describing is actually the bug we are having atm.

akatsoulas:
r+

PR #1487, opened 2016-07-20
Remove referral_source for django browserid.

akatsoulas:


akatsoulas:
@johngian r?

johngian:
Code looks OK. I also tested the profile creation locally and works fine.
r+

PR #1491, opened 2016-07-25
Export data from /admin asynchronously

johngian:


johngian:
@akatsoulas r?

mozillians/common/mixins.py
```diff
@@ -0,0 +1,63 @@
...
+        context = {}
+        context.update(self.admin_site.each_context(request))
+
+        context['form'] = form
+        context['opts'] = self.model._meta
+        request.current_app = self.admin_site.name
+        return TemplateResponse(request, [self.export_template_name], context)
+
+
+MozilliansAdminExportMixin = import_string(settings.ADMIN_EXPORT_MIXIN)
```
akatsoulas: Add a docstring/comment to explain what you do here.

mozillians/common/mixins.py
```diff
@@ -0,0 +1,63 @@
...
+from django.utils.module_loading import import_string
+
+import boto
+
+from celery.task import task
+from import_export.admin import ExportMixin
+from import_export.forms import ExportForm
+
+
+@task(soft_time_limit=10 * 60)
```
akatsoulas: Add a flag in the settings page for this.

mozillians/users/admin.py
```diff
@@ -155,9 +155,9 @@ class DateJoinedFilter(SimpleListFilter):
     parameter_name = 'date_joined'
 
     def lookups(self, request, model_admin):
-
-        return map(lambda x: (str(x.year), x.year),
-                   User.objects.datetimes('date_joined', 'year'))
+        join_dates = User.objects.values_list('date_joined', flat=True)
+        join_years = map(lambda x: x.year, join_dates)
+        return map(lambda x: (str(x), x), set(join_years))
```
glogiotatidis: nit: since you're touching this, list comprehension is more pythonic these days.

mozillians/common/tests/test_mixins.py
```diff
@@ -0,0 +1,51 @@
...
+
+        with self.login(user) as client:
+            response = client.get(export_url)
+            eq_(response.status_code, 200)
+
+    @override_settings(AWS_ACCESS_KEY_ID='foo')
+    @override_settings(AWS_SECRET_ACCESS_KEY='bar')
+    @override_settings(MOZILLIANS_ADMIN_BUCKET='s3-bucket')
+    @patch('mozillians.common.mixins.S3ExportMixin.get_export_filename')
+    @patch('mozillians.common.mixins.boto')
```
glogiotatidis: I'd generally avoid patching the whole test and instead wrap with `with` the minimum code that needs to be patched. Not wrong but it can help you avoid weird test results if lets say some time in the future group creation uses boto for something.

mozillians/common/mixins.py
```diff
@@ -0,0 +1,67 @@
...
+        context.update(self.admin_site.each_context(request))
+
+        context['form'] = form
+        context['opts'] = self.model._meta
+        request.current_app = self.admin_site.name
+        return TemplateResponse(request, [self.export_template_name], context)
+
+
+# Allow configuring admin export mixin in project settings
+MozilliansAdminExportMixin = import_string(settings.ADMIN_EXPORT_MIXIN)
```
glogiotatidis: Plan to use something else as well?

johngian: I added this so we can easily fall back to the default one in case something goes wrong. I guess a comment would be helpful :)

mozillians/common/tests/test_mixins.py
```diff
@@ -0,0 +1,51 @@
...
+        user = UserFactory.create(is_superuser=True, is_staff=True)
+        export_url = reverse('admin:users_userprofile_export')
+
+        # NDA group required for admin response rendering
+        GroupFactory.create(name=settings.NDA_GROUP)
+
+        mock_key = MagicMock()
+        mock_bucket = MagicMock()
+        mock_connection = MagicMock()
+        mock_bucket.new_key.return_value = mock_key
```
glogiotatidis: I believe that's not needed as you don't use `mock_key` anywhere and mock will return a mock anyway.

mozillians/common/mixins.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def export_action(self, request, *args, **kwargs):
+        formats = self.get_export_formats()
+        form = ExportForm(formats, request.POST or None)
+
+        if form.is_valid():
+            file_format = formats[int(form.cleaned_data['file_format'])]()
+            queryset = self.get_export_queryset(request)
+            task_id = self.get_export_data(file_format, queryset)
+            msg = 'Data export task spawned with id: {}'.format(task_id)
```
glogiotatidis: Maybe useful to include a link or a least the name of the saved file?

glogiotatidis:
Good job! r+ with some nits only :)

mozillians/common/tests/test_mixins.py
johngian: We are using this in the assertions at the end of the test. Without this, the test is failing.

PR #1492, opened 2016-07-25
[fix bug 1289114] Do not subscribe a registered user multiple times.

akatsoulas:


mozillians/users/tasks.py
```diff
@@ -130,37 +130,30 @@ def update_basket_task(instance_id, newsletters=[]):
...
-            # FIXME: We should also remove their previous phonebook record from Exact Target, but
-            # basket doesn't have a custom API to do that. (basket never deletes anything.)
-
             # That was all successful. Update the token.
-            instance.basket_token = subscribe_result['token']
             token = subscribe_result['token']
             # Don't call .save() on a userprofile from here, it would invoke us again
             # via the post-save signal, which would be pointless.
             UserProfile.objects.filter(pk=instance.pk).update(basket_token=token)
+        else:
```
johngian: This else case is wrong here. There are cases where user has basket token but needs to get subscribed to another newsletter (eg. vouched user that is getting subscribed to NDA)

akatsoulas: It was an indentation error

johngian:
Code looks OK.

johngian:
r+ with tests passing

PR #1493, opened 2016-07-26
[Fix bug 1289117] Add entry in UserProfile to tag profile as spam.

johngian:
* Expose spam bit in /admin
* Allow filtering based on `is_spam` value


johngian:
@akatsoulas r?

akatsoulas:
r+

PR #1494, opened 2016-07-28
[Fix bug 1289118] Add helper method to check for spam using Akismet API

johngian:


mozillians/common/utils.py
```diff
@@ -16,3 +18,25 @@ def absolutify(url):
...
+
+    parameters = {
+        'blog': settings.SITE_URL,
+        'user_ip': user_ip,
+        'user_agent': user_agent,
+    }
+
+    parameters.update(optional)
+
+    url = 'https://{0}.rest.akismet.com/1.1/comment-check'.format(settings.AKISMET_API_KEY)
```
akatsoulas: Maybe it's better to use a global variable to the top of the file for this.

akatsoulas:
r+ with tests passing

PR #1499, opened 2016-08-02
Simplify update_basket task.

akatsoulas:


johngian:
Please fix typo in commit message/PR title

mozillians/users/tasks.py
```diff
@@ -102,18 +103,13 @@ def update_basket_task(instance_id, newsletters=[]):
...
-            except (MaxRetriesExceededError, basket.BasketException):
-                msg = exception.message
-                _email_basket_managers('update_phonebook', token, msg)
-            return
+            retry = True
 
         # Check if the users needs to be subscribed to additional newsletters.
         newsletters_to_subscribe = [nl for nl in newsletters if nl not in result['newsletters']]
 
         old_email = result['email']
```
johngian: When ``retry==True`` ``result`` var is undefined.

johngian:
Instead of having these endless ``if/else/elif`` cases i suggest we wrap all the basket calls to a helper method that takes as arguments:

* basket method
* basket method args
* task

and does the heavy-lifting of celery retries.

akatsoulas:
Closing this as it's not built in the right direction. I will open a new one for this.

PR #1500, opened 2016-08-05
[fix bug 1292358] Replace 'icontains' with 'startswith' in skills que…

akatsoulas:
…ries.


mozillians/groups/views.py
```diff
@@ -490,7 +490,7 @@ def get_queryset(self):
 
         qs = Skill.objects.all()
         if self.q:
-            qs = qs.filter(name__icontains=self.q)
+            qs = qs.filter(name__startswith=self.q)
```
johngian: Please use ``istartswith``

johngian:
r+wc

PR #1501, opened 2016-08-17
[Fix bug 1289119] Create celery task to check for spam

johngian:


johngian:
@akatsoulas r?

mozillians/users/tasks.py
```diff
@@ -282,3 +283,13 @@ def remove_incomplete_accounts(days=INCOMPLETE_ACC_MAX_DAYS):
 def check_celery():
     """Dummy celery task to check that everything runs smoothly."""
     pass
+
+
+@task
```
akatsoulas: Is this going to be a periodic task? If yes, I would suggest to schedule it here. When pr #1483 is merged we won't have the ability to do so from the admin panel.

mozillians/common/utils.py
```diff
@@ -26,6 +23,13 @@ def absolutify(url):
 def akismet_spam_check(user_ip, user_agent, **optional):
     """Checks for spam content against Akismet API."""
 
+    AKISMET_API_KEY = getattr(settings, 'AKISMET_API_KEY', '')
+
+    if not AKISMET_API_KEY:
+        return None
+
+    AKISMET_URL = 'https://{0}.rest.akismet.com/1.1/comment-check'.format(AKISMET_API_KEY)
```
akatsoulas: Why did you move this here?

mozillians/users/tasks.py
johngian: It's not a periodic task. It gets triggered on each profile save.

mozillians/phonebook/views.py
```diff
@@ -238,6 +239,19 @@ def get_request_data(form):
             for f in curr_forms:
                 f.save()
 
+            # Spawn task to check for spam
```
akatsoulas: Should we spawn this task only when the bio field is populated with data to avoid a few calculations?

johngian: We are checking against many fields and not just bio. Akismet might be able to check for spam even by just providing ``REMOTE_ADDR``.

akatsoulas:
r+wc

PR #1502, opened 2016-08-24
[fix bug 1262853] Add recaptcha to profile registration.

akatsoulas:


akatsoulas:
@johngian r?

johngian:
Please fix ``flake8`` issues.

johngian:
I would prefer to reduce the spacing here.

selection_044

The following looked better locally:
```
#privacy-verify {
    margin-bottom: 0px;
}
```

johngian:
UI is broken in mobile view:

selection_045

johngian:
Label ``recaptcha widget`` on hover looks kind of weird. I suggest we disable it for this element specifically.

mozillians/phonebook/tests/test_profile_edit.py
```diff
@@ -160,8 +161,9 @@ def test_succesful_registration(self, info_mock):
             'lat': 40.005814,
             'lng': -3.42071,
             'optin': True,
-            'registration_section': ''
+            'registration_section': '',
         }
+        mocked_form.is_valid.return_value = True
```
johngian: This causes the form to be always valid without checking the data.

akatsoulas: This test is supposed to test for the successful creation of a profile by feeding the data to the view and not to check the form for validation errors. I am going to refactor this test to assert for a successful profile  but I don't feel that mocking the form is a problem here

mozillians/static/mozillians/css/main.less
```diff
@@ -2650,3 +2650,18 @@ fieldset {
...
     }
 }
+
+/* Recaptcha
+============ */
+#privacy-verify {
+    margin-bottom: 0px;
+}
+
+@media screen and (max-height: 575px) {
```
johngian: I would prefer to be consistent with the media queries in the rest of the project. This works fine locally:
```
@media (max-width: @breakTablet)
```

johngian:
r+wc with tests passing

PR #1504, opened 2016-09-08
Add management plugin in RabbitMQ for Docker.

akatsoulas:


docker-compose.yml
```diff
@@ -10,7 +10,9 @@ es:
 memcached:
   image: memcached
 broker:
-  image: rabbitmq
+  image: rabbitmq:3-management
+  ports:
+    - "15672:15672"
```
johngian: I think that this entry here should be:

```
ports:
  - "8080:15672"
```

johngian:
r+

PR #1506, opened 2016-09-09
[bug 1289117] Make 'is_spam' field writable in the admin interface.

akatsoulas:


akatsoulas:
@johngian r?

Really minor but it would be helpful to be able to toggle this bit (missed profiles or false positives).

johngian:
r+

PR #1510, opened 2016-09-12
Fix async export serialization

johngian:


johngian:
@akatsoulas r?

mozillians/common/mixins.py
```diff
@@ -15,14 +16,25 @@
...
+
     data = resource_class().export(queryset)
     export_data = file_format.export_data(data)
 
     # Store file to AWS S3
-    conn = boto.connect_s3(settings.AWS_ACCESS_KEY_ID, settings.AWS_SECRET_ACCESS_KEY)
+    kwargs = {
+        'aws_access_key_id': settings.AWS_ACCESS_KEY_ID,
+        'aws_secret_access_key': settings.AWS_SECRET_ACCESS_KEY,
+        'calling_format': OrdinaryCallingFormat()
```
akatsoulas: Please add a couple of lines in the comment above explaining why you are using this. Probably this will save us some time in the future.

akatsoulas:
Tests need some fine tuning.

mozillians/common/mixins.py
```diff
@@ -15,9 +15,15 @@
...
 
 @task(soft_time_limit=ADMIN_EXPORT_TIMEOUT)
-def async_data_export(file_format, resource_class, values_list, model, filename):
+def async_data_export(file_format, values_list, qs_model, filename):
     """Task to export data from admin site and store it to S3."""
-    queryset = model.objects.filter(id__in=values_list)
+
+    from django.contrib import admin
+
+    admin_obj = admin.site._registry[qs_model]
```
akatsoulas: You can pass this object directly from the `get_export_data` method in `S3ExportMixin` class and avoid using the `_registry`.

mozillians/common/mixins.py
```diff
@@ -15,14 +16,25 @@
...
```
akatsoulas: You can pass this object directly from the `get_export_data` method in `S3ExportMixin` class and avoid using the `_registry`.

akatsoulas:
r+wc 🍺

mozillians/common/mixins.py
johngian: There are some edge cases with the way resource classes are defined in django import export. I suggest we tackle our current issue this way and revisit it in the future.

PR #1515, opened 2016-09-15
Allow users to report abusive/spam content

johngian:


mozillians/jinja2/phonebook/profile.html
```diff
@@ -129,6 +129,39 @@
...
 
+      {% if abuse_form %}
+        <section id="profile-actions">
+          <div class="actions-details">
+            <button class="btn btn-default btn-xs" type="submit" data-toggle="modal" data-target="#ReportModal">Report</button>
+          </div>
+        </section>
+
+        <!-- Report Modal -->
+        <div class="modal fade" id="ReportModal" tabindex="-1" role="dialog" aria-labelledby="ReportModalLabel">
```
akatsoulas: Maybe it's a good idea to hide this field for vouched users.

mozillians/phonebook/views.py
```diff
@@ -128,6 +130,15 @@ def view_profile(request, username):
             profile.set_instance_privacy_level(
                 request.user.userprofile.privacy_level)
 
+        if request.user.is_authenticated() and request.user.userprofile.is_vouched:
+            abuse_report = AbuseReport(reporter=request.user.userprofile, profile=profile)
+            abuse_form = forms.AbuseReportForm(request.POST or None, instance=abuse_report)
+            if abuse_form.is_valid():
```
akatsoulas: Please make sure that the profile reported here, is not a vouched one

mozillians/phonebook/forms.py
```diff
@@ -428,3 +429,16 @@ class APIKeyRequestForm(happyforms.ModelForm):
...
+class AbuseReportForm(happyforms.ModelForm):
+
+    class Meta:
+        model = AbuseReport
+        fields = ('type',)
+        widgets = {
+            'type': RadioSelect
+        }
+        labels = {
+            'type': _(u'What would you like to report?')
```
akatsoulas: I don't have a good suggestion atm but I would prefer something different here.

mozillians/phonebook/views.py
johngian: Why not. Even vouched accounts can post inappropriate content. I don't see what's the value on restricting the logic here.

mozillians/phonebook/views.py
```diff
@@ -240,7 +252,7 @@ def get_request_data(form):
                 f.save()
 
             # Spawn task to check for spam
-            if not profile.is_spam:
+            if not profile.is_vouched:
```
johngian: After a quick discussion we decided that its safer to exclude only users that have the ``can_vouch`` flag.

mozillians/phonebook/forms.py
johngian: Let me know if you have a specific suggestion. Else i suggest we revisit the phrasing after we merge this.

mozillians/users/models.py
```diff
@@ -703,6 +702,23 @@ def update_vouch_flags(sender, instance, **kwargs):
     profile.save(**{'autovouch': False})
 
 
+class AbuseReport(models.Model):
+    TYPE_SPAM = 'spam'
+    TYPE_INAPPROPRIATE = 'inappropriate'
```
johngian: I suggest we leave the var name as it is for consistency with the UI.

mozillians/phonebook/views.py
akatsoulas: That's maybe true for inappropriate but this argument does not hold for spam.

mozillians/jinja2/phonebook/profile.html
johngian: Tackled in the ``view`` level.

mozillians/phonebook/views.py
johngian: Let's restrict this to people that ``can_vouch``.

mozillians/users/migrations/0010_auto_20160915_0905.py
```diff
@@ -0,0 +1,29 @@
...
+            'profile': profile,
+            'is_akismet': True
+        }
+        AbuseReport.objects.create(**kwargs)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('users', '0009_abusereport'),
```
akatsoulas: Backwards method please!

johngian:
@akatsoulas PR updated

mozillians/phonebook/views.py
```diff
@@ -128,6 +130,16 @@ def view_profile(request, username):
             profile.set_instance_privacy_level(
                 request.user.userprofile.privacy_level)
 
+        if (request.user.is_authenticated() and request.user.userprofile.is_vouched and
+                not profile.can_vouch):
+            abuse_report = AbuseReport(reporter=request.user.userprofile, profile=profile)
+            abuse_form = forms.AbuseReportForm(request.POST or None, instance=abuse_report)
+            if abuse_form.is_valid():
```
akatsoulas: There is no check to prevent me to report a user multiple times. Probably it's a good idea to allow one report per profile per user. Thoughts?

akatsoulas:
r+ 🚢 it

PR #1520, opened 2016-09-22
[Fix bug 1304728] Disable recaptcha check on dev/stage.

johngian:


akatsoulas:
After discussing this with @johngian, we are going with this solution [0] and amend the selenium tests to select the captcha field.

[0] https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha-v2-what-should-i-do

PR #1523, opened 2016-09-26
[fix bug 1304683] Refactor update_basket_task.

akatsoulas:


mozillians/users/tasks.py
```diff
@@ -63,22 +65,130 @@ def _email_basket_managers(action, email, error_message):
...
+
+        exception: {exc}
+
+        args: {args}
+
+        kwargs: {kwargs}
+
+        exception info: {einfo}
+
+        %s
```
johngian: Do we need this ``%s``?

mozillians/users/tasks.py
```diff
@@ -63,22 +65,130 @@ def _email_basket_managers(action, email, error_message):
...
+        raise self.retry(exc=exc)
+    return result
+
+
+@shared_task(bind=True, base=DebugBasketTask, default_retry_delay=BASKET_TASK_RETRY_DELAY,
+             max_retries=BASKET_TASK_MAX_RETRIES)
+def subscribe_user_task(self, result, profile_id, email='', newsletters=[]):
+    """Subscribes a user to basket newsletters.
+
+    Also updates the email registered to basket based on the query
```
johngian: I think that this is kind of misleading since no lookup nor email change is happening in this task.

mozillians/users/tasks.py
```diff
@@ -63,22 +65,130 @@ def _email_basket_managers(action, email, error_message):
...
+        return subscribe_result
+    return {'result': 'not subscribed'}
+
+
+@shared_task(bind=True, base=DebugBasketTask, default_retry_delay=BASKET_TASK_RETRY_DELAY,
+             max_retries=BASKET_TASK_MAX_RETRIES)
+def unsubscribe_user_task(self, result, profile_id=None, email='', newsletters=[], optout=True):
+    """Removes a user from the Basket subscription."""
+
+    if not email and not profile_id:
```
johngian: What's the case that there is no ``email`` and no ``profile_id``?

mozillians/users/tasks.py
```diff
@@ -63,22 +65,130 @@ def _email_basket_managers(action, email, error_message):
...
+        newsletters_to_unsubscribe = [nl for nl in MOZILLIANS_NEWSLETTERS
+                                      if nl in result['newsletters']]
+    else:
+        newsletters_to_unsubscribe = [nl for nl in newsletters if nl in result['newsletters']]
+
+    # Unsubscribe the calculated newsletters
+    try:
+        unsubscribe_result = basket.unsubscribe(email=email,
+                                                newsletters=newsletters_to_unsubscribe,
+                                                optout=optout)
```
johngian: If ``optout`` is True user is going to be unsubscribed from all newsletters regardless of the ``newsletter`` value.

mozillians/users/tasks.py
```diff
@@ -63,22 +65,130 @@ def _email_basket_managers(action, email, error_message):
...
+        email = profile.user.email
+
+    newsletters_to_subscribe = [nl for nl in newsletters if nl not in result['newsletters']]
+
+    if newsletters_to_subscribe:
+        try:
+            subscribe_result = basket.subscribe(email,
+                                                newsletters_to_subscribe,
+                                                sync='Y',
+                                                trigger_welcome='N')
```
johngian: Do we need to ``trigger_welcome`` email on email change?

mozillians/users/tasks.py
johngian: This needs also code changes in the line where we spawn the task.

mozillians/phonebook/views.py
```diff
@@ -341,6 +341,8 @@ def change_primary_email(request, email_pk):
     with transaction.atomic():
         user.save()
         alternate_email.save()
+    # Notify Basket about this change
+    update_email_in_basket.delay(profile.id, primary_email, alternate_email.identifier)
```
johngian: Also what happens when the change of email is applied from admin? We've had a lot of cases where users lost access to their emails and required manual changes from /admin.

mozillians/users/tasks.py
```diff
@@ -63,22 +65,130 @@ def _email_basket_managers(action, email, error_message):
...
+        except MaxRetriesExceededError as exc:
+            raise exc
+        except basket.BasketException as exc:
+            self.retry(exc=exc)
+        return subscribe_result
+    return {'result': 'not subscribed'}
```
johngian: Since we are not using it somewhere i suggest this to ``return None``.

mozillians/users/tasks.py
johngian: We can always use:
```
if result
     foo
```

mozillians/users/tasks.py
```diff
@@ -63,22 +65,130 @@ def _email_basket_managers(action, email, error_message):
...
+    except basket.BasketException:
+        self.retry(exc=exc)
+    return unsubscribe_result
+
+
+@shared_task()
+def subscribe_user_to_basket(instance_id, newsletters=[]):
+    """Subscribe a user to Basket.
 
     This task subscribes a user to Basket, if not already subscribed
```
johngian: This docstring seems outdated.

mozillians/users/tasks.py
```diff
@@ -89,80 +199,37 @@ def update_basket_task(instance_id, newsletters=[]):
...
+    try:
+        instance = UserProfile.objects.get(pk=instance_id)
+    except UserProfile.DoesNotExist:
+        instance = None
+
+    if not BASKET_ENABLED or not instance or not waffle.switch_is_active('BASKET_SWITCH_ENABLED'):
+        return
+
+    lookup_old_email_subtask = lookup_user_task.subtask((instance_id, old_email),)
+    unsubscribe_subtask = unsubscribe_user_task.subtask()
```
johngian: Please add a comment here to mention that the only arg being used is the only from the chained results.

mozillians/users/tasks.py
```diff
@@ -89,80 +199,37 @@ def update_basket_task(instance_id, newsletters=[]):
...
+    # we don't care for the result of the previous subtask (unsubscribe), thus the si
+    lookup_new_email_subtask = lookup_user_task.si(instance_id, new_email)
+    subscribe_subtask = subscribe_user_task.subtask((instance_id, new_email,))
```
johngian: Here we are loosing the ``newsletters`` argument from the initial lookup.

mozillians/users/tasks.py
akatsoulas: Nope.

akatsoulas:
@johngian pr updated. I changed a few things compared to the initial one so some comments are not relevant any more.

mozillians/users/tasks.py
```diff
@@ -89,80 +194,26 @@ def update_basket_task(instance_id, newsletters=[]):
...
-        # via the post-save signal, which would be pointless.
-        UserProfile.objects.filter(pk=instance.pk).update(basket_token=retval['token'])
+    This task is triggered when users change their email.
+    The task checks whether the user is already subscribed with the old email.
+    If there is a subscription we remove it and we register the new email.
+    """
+    if not BASKET_ENABLED or not waffle.switch_is_active('BASKET_SWITCH_ENABLED'):
+        return
+
+    group(
```
johngian: What if the second group is scheduled first and the user is unsubscribed?

johngian:
Please add some comment/docstring that explains that not all the tasks should be used directly but use the top-level tasks instead.

mozillians/users/tasks.py
```diff
@@ -63,23 +62,129 @@ def _email_basket_managers(action, email, error_message):
...
+    The email subscribed is provided either from the result of the lookup task
+    or from the profile of the user.
+    """
+
+    if not result and not email:
+        return None
+
+    newsletters_to_subscribe = []
+    if result['status'] == 'ok':
+        if not email:
```
johngian: Please add a comment that explains how we are facilitating both ``results`` and ``email`` arg.

mozillians/users/tasks.py
```diff
@@ -63,23 +62,129 @@ def _email_basket_managers(action, email, error_message):
...
+        return None
+
+    newsletters_to_unsubscribe = []
+    if result['status'] == 'ok':
+        email = result.get('email')
+        token = result.get('token')
+
+        # only unsubscribe from our newsletters
+        if not newsletters:
+            newsletters_to_unsubscribe = [nl for nl in MOZILLIANS_NEWSLETTERS
```
johngian: Please be consistent between here and the previous similar functionality. Either use list comprehension or sets.

mozillians/users/tasks.py
akatsoulas: This is a valid point. I just updated the code to make sure that both subscribe and unsubscribe task will get the correct results from the lookup task.

mozillians/users/tasks.py
```diff
@@ -152,13 +154,13 @@ def unsubscribe_user_task(self, result, newsletters=[], optout=False):
     if result['status'] == 'ok':
         email = result.get('email')
         token = result.get('token')
+        nwsltrs = result.get('newsletters')
```
johngian: nit: i would go for ``curr_newsletters`` instead to be more descriptive

johngian:
nit: Please update typo in commit message

johngian:
I drafted some tests here:
https://github.com/mozilla/mozillians/commit/23324ae11729969526aebed3fdb556eb96199556

akatsoulas:
@johngian I believe this is good to go. I cherry-picked your commit and added one more with a couple more tests and a few minor enhancements.

PR #1526, opened 2016-09-27
[bug 1302721] Add django-autocomplete light to ReportAbuse model in a…

akatsoulas:
…dmin.

### DIFF
--- a/mozillians/users/admin.py
+++ b/mozillians/users/admin.py
@@ -654,7 +654,19 @@ class VouchAdmin(admin.ModelAdmin):
 admin.site.register(Vouch, VouchAdmin)
 
 
+class AbuseReportAutocompleteForm(forms.ModelForm):
+
+    class Meta:
+        model = AbuseReport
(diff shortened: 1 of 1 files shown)


johngian:
@akatsoulas ``flake8`` is failing

mozillians/users/admin.py
```diff
@@ -654,7 +654,19 @@ class VouchAdmin(admin.ModelAdmin):
...
 
 
+class AbuseReportAutocompleteForm(forms.ModelForm):
+
+    class Meta:
+        model = AbuseReport
+        fields = '__all__'
+        widgets = {
+            'profile': autocomplete.ModelSelect2(url='users:vouchee-autocomplete'),
+            'reporter': autocomplete.ModelSelect2(url='users:voucher-autocomplete'),
```
johngian: I suggest we use the same autocomplete endpoint for both ``profile`` and ``reporter``. It's in admin so its better to have less restrictions.

PR #1528, opened 2016-09-28
[fix bug 1305939] Fix UserProfile query in spam celery task.

akatsoulas:


johngian:
@akatsoulas tests are failing

mozillians/users/tasks.py
```diff
@@ -14,6 +14,7 @@
 from elasticutils.utils import chunked
 
 from mozillians.common.utils import akismet_spam_check
+from mozillians.common.templatetags import get_object_or_none
```
johngian: This should be:
``from mozillians.common.templatetags.helpers import get_object_or_none``

akatsoulas:
@johngian pr updated

PR #1529, opened 2016-10-03
Upgrade Celery to version 3.1.24.

akatsoulas:


johngian:
Tests are passing so there shouldn't be an issue with pip dependencies.
r+

PR #1531, opened 2016-10-06
Update CSP directives to fix CSP violations.

johngian:


johngian:
CSP violation reports that we are fixing (from sentry):

```
violated-directive: default-src https://mozillians.org:443 https://*.mapbox.com:443 https://*.persona.org:443

violated-directive: default-src https://mozillians.org https://*.mapbox.com https://*.persona.org
```

johngian:
Also from the documentation:

```
Alternatively, please add the following values to the directives:

    script-src https://www.google.com/recaptcha/, https://www.gstatic.com/recaptcha/
    frame-src https://www.google.com/recaptcha/
    style-src 'unsafe-inline'
```

akatsoulas:
Let's test this with only the frame directive. r+wc

akatsoulas:
r+

PR #1536, opened 2016-10-13
Remove unused scripts.

akatsoulas:


johngian:
I don't think that we are using ``bin/jenkins.sh`` maybe we should consider removing the entire file.

PR #1548, opened 2016-11-11
Deprecate BrowserID support

johngian:


glogiotatidis:
good luck and godspeed!

giphy

johngian:
@akatsoulas r?

I tested the following locally:

* Login as non-existing user
  * Prompts for new profile creation
* Login as existing user
  * Redirects to profile
* Login with alternate email
  * Authenticate user with matching alternate email

mozillians/common/authbackend.py
```diff
@@ -4,16 +4,9 @@
...
-from django.core.urlresolvers import reverse
-from django.db import IntegrityError
 from django.db.models import Q
 
-from django_browserid.auth import BrowserIDBackend
-from django_browserid.base import RemoteVerifier, get_audience
-from django_browserid.http import JSONResponse
-from django_browserid.views import Verify
-from django.utils.translation import ugettext as _
+from mozilla_django_oidc.auth import OIDCAuthenticationBackend
```
akatsoulas: Do we need the `calculate_username` method? Can we use the username_algo from the odic lib?

mozillians/common/authbackend.py
```diff
@@ -39,83 +32,14 @@ def calculate_username(email):
...
-
-        return user
-
-    def filter_users_by_email(self, email):
         account_type = ExternalAccount.TYPE_EMAIL
         alternate_emails = ExternalAccount.objects.filter(type=account_type, identifier=email)
         primary_email_qs = Q(email=email)
         alternate_email_qs = Q(userprofile__externalaccount=alternate_emails)
-        return self.User.objects.filter(primary_email_qs | alternate_email_qs).distinct()
+        return self.UserModel.objects.filter(primary_email_qs | alternate_email_qs).distinct()
```
akatsoulas: Can't use just use something like:

```python
q = Q(call_super_filter_users)
return self.UserModel.objects.filter(Q(alternate) | Q(super))
```

mozillians/common/middleware.py
```diff
@@ -94,6 +94,10 @@ def _is_lang_change(self, request):
         return not any(request.path.endswith(url) for url in self.exempt_urls)
 
     def process_request(self, request):
+        for view_url in self.exempt_urls:
```
akatsoulas: Please add a comment here.

mozillians/common/templatetags/helpers.py
```diff
@@ -284,8 +285,13 @@ def url(viewname, *args, **kwargs):
     return reverse(viewname, args=args, kwargs=kwargs)
 
 
+@library.global_function
+def nonprefixed_url(viewname, *args, **kwargs):
```
akatsoulas: Comments please.

mozillians/jinja2/phonebook/betasearch.html
```diff
@@ -16,11 +16,7 @@
     <p class="alert">
       {{ _('You are currently searching public profiles.') }}
       {% if not user.is_authenticated() %}
-        {% trans next=request.get_full_path(), title=_('Sign In') %}
-          View more results by
-          <a href="#" class="browserid-login" title="{{ title }}" data-next="{{ next }}">
-            logging in</a>.
-        {% endtrans %}
+          View more results by logging in.
```
akatsoulas: Shouldn't we include the login link here?

mozillians/jinja2/phonebook/includes/profile_edit_basic.html
```diff
@@ -113,41 +113,33 @@
           </div>
         </fieldset>
 
-        <fieldset id="alternate-email">
-          <label class="email-label">{{ _('Secondary e-mail addresses') }}</label>
-          {{ alternate_email_formset.management_form }}
+        {% if alternate_email_formset.get_queryset() %}
```
akatsoulas: Why is this `if` needed?

mozillians/jinja2/phonebook/search.html
```diff
@@ -16,11 +16,7 @@
     <p class="alert">
       {{ _('You are currently searching public profiles.') }}
       {% if not user.is_authenticated() %}
-        {% trans next=request.get_full_path(), title=_('Sign In') %}
-          View more results by
-          <a href="#" class="browserid-login" title="{{ title }}" data-next="{{ next }}">
-            logging in</a>.
-        {% endtrans %}
+        View more results by logging in.
```
akatsoulas: Same comment here about the missing link.

mozillians/settings/base.py
```diff
@@ -273,8 +270,8 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
 
 # Authentication settings
 AUTHENTICATION_BACKENDS = (
+    'mozillians.common.authbackend.MozilliansAuthBackend',
     'django.contrib.auth.backends.ModelBackend',
-    'mozillians.common.authbackend.MozilliansAuthBackend'
```
akatsoulas: Why are you changing this? If the order is important, please add a comment.

akatsoulas:
@glogiotatidis what about a quick pass?

mozillians/common/authbackend.py
johngian: Yes, we are using it to define usernames on registration and we want to keep it for consistency with old users. This is the algo that we pass to the OIDC lib.

mozillians/common/authbackend.py
johngian: No, super returns a queryset.

mozillians/jinja2/phonebook/includes/profile_edit_basic.html
johngian: At the moment we are not migrating alternate email functionality. That means that we need to hide the alternate email section if there are no emails in the queryset.

mozillians/common/authbackend.py
akatsoulas: yeah obviously... Q works just fine or this could be reduced down to `my_or_q = (super_queryset | other_queryset).distinct()
disclaimer: not tested but seems fun

mozillians/common/templatetags/helpers.py
johngian: Fixed in next line.

mozillians/common/middleware.py
```diff
@@ -94,6 +94,11 @@ def _is_lang_change(self, request):
         return not any(request.path.endswith(url) for url in self.exempt_urls)
 
     def process_request(self, request):
+        # Don't apply middleware to requests matching exempt URLs
+        for view_url in self.exempt_urls:
+            if re.match(view_url, request.path):
```
akatsoulas: Please use something like:
```python
if view_url in self.exempt_urls:
    return None
```

It's probably much faster than a regex.

mozillians/common/templatetags/helpers.py
akatsoulas: Why don't you just use the `url` tag?

mozillians/common/middleware.py
johngian: But doesn't do the same thing.

mozillians/common/templatetags/helpers.py
johngian: We are overriding ``url`` to add the locale prefix. That breaks the OIDC urls.

mozillians/settings/base.py
```diff
@@ -583,3 +557,15 @@ def _browserid_audiences():
 # Recaptcha
 NORECAPTCHA_SITE_KEY = 'site_key'
 NORECAPTCHA_SECRET_KEY = 'secret_key'
+
+
+# Django OIDC
+
+
```
akatsoulas: Too many blank lines here.

PR #1558, opened 2016-12-01
Multiple L10N fixes

johngian:
* Exclude /api and /admin URLs from L10N middleware.
* Fix LANGUAGES definition format.
* Use regex for URLs that are not to be localized.

### DIFF
--- a/mozillians/settings/base.py
+++ b/mozillians/settings/base.py
@@ -72,8 +72,11 @@ STANDALONE_DOMAINS = [TEXT_DOMAIN, 'djangojs']
 LANGUAGE_CODE = 'en-US'
 LOCALE_PATHS = [path('locale')]
 EXEMPT_L10N_URLS = [
-    '/oidc/authenticate/',
-    '/oidc/callback/'
+    '^/oidc/authenticate/',
+    '^/oidc/callback/',
+    '^/api/v1/',
+    '^/api/v2/',
+    '^/admin/'
 ]
 
 # Tells the extract script what files to parse for strings and what functions to use.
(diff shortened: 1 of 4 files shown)


mozillians/settings/base.py
```diff
@@ -72,8 +72,10 @@
 LANGUAGE_CODE = 'en-US'
 LOCALE_PATHS = [path('locale')]
 EXEMPT_L10N_URLS = [
-    '/oidc/authenticate/',
-    '/oidc/callback/'
+    '^/oidc/authenticate/',
+    '^/oidc/callback/',
+    '^/api/v2/',
+    '^/admin/'
```
akatsoulas: Let's add `/api/v1` here too.

johngian:
We should also fix LANGUAGES references that handle languages as dict.

mozillians/settings/base.py
```diff
@@ -118,11 +120,11 @@ def get_langs():
 def lazy_langs():
     from product_details import product_details
 
-    return dict([(lang.lower(), product_details.languages[lang]['native'])
-                 for lang in get_langs() if lang in product_details.languages])
+    return [(lang.lower(), product_details.languages[lang]['native'])
```
akatsoulas: Out of curiosity, what's the reason for this change?  If this is not something relevant to this PR, please revert the change.

johngian: We are changing this because of the way ``Puente`` works. It tries to get closer to how django internal l10n support works.

https://docs.djangoproject.com/en/1.10/ref/settings/#languages

PR #1561, opened 2016-12-18
[fix bug 1324272] Do not submit empty search form

MikkCZ:


MikkCZ:
@johngian, @akatsoulas r?

PR #1566, opened 2017-01-04
Re-add multiple emails in mozillians.org

akatsoulas:


mozillians/common/authbackend.py
```diff
@@ -33,13 +33,27 @@ def calculate_username(email):
...
         alternate_email_qs = Q(userprofile__externalaccount=alternate_emails)
-        return self.UserModel.objects.filter(primary_email_qs | alternate_email_qs).distinct()
+        user_q = self.UserModel.objects.filter(primary_email_qs | alternate_email_qs).distinct()
+
+        # In this case we have a registered user who is adding a secondary email
+        if request_user.is_authenticated() and not user_q:
+            ExternalAccount.objects.create(type=account_type,
+                                           user=request_user.userprofile,
+                                           identifier=email)
+            return [request_user]
```
johngian: I would prefer to return a qs instead of a list with an object.

mozillians/common/tests/test_authbackend.py
```diff
@@ -0,0 +1,71 @@
...
+        ok_(not email_q.exists())
+        returned_user = self.backend.filter_users_by_claims(claims)
+        email_q = ExternalAccount.objects.filter(type=ExternalAccount.TYPE_EMAIL,
+                                                 user=user.userprofile,
+                                                 identifier='bar@example.com')
+        ok_(email_q.exists())
+        eq_(len(returned_user), 1)
+        eq_(returned_user[0], user)
+
+    def test_email_already_exists(self):
```
johngian: Does it make sense to have the same test but using the primary email instead? To test both cases.

mozillians/jinja2/phonebook/includes/profile_edit_basic.html
```diff
@@ -139,6 +137,14 @@
 
             </div>
           {% endfor %}
+
+          <!-- Add a new email -->
+          <div id="add-email">
+            <a id="nav-login" title="{{_('Add email using Auth0')}}"
```
johngian: Since users don't really know/care that we are using auth0 (its mozilla branded anyway) I suggest we use something like ``Add email``.

johngian:
tested locally:

* Logging in with primary email
* Logging in with alternate email
* Add alternate email
* Make alternate email primary

mozillians/common/authbackend.py
akatsoulas: Since we already have the object, I would prefer not to hit the db just to create a QS

PR #1571, opened 2017-01-05
Remove migrations debug code.

johngian:


akatsoulas:
I would suggest to leave this there. It's not a bad idea to have some verbosity in our logs.

PR #1575, opened 2017-01-09
[bug 1254083][regression] Invalidate map size.

akatsoulas:


johngian:
Tested it locally

* Fix works fine
* There is a small glitch the first time you click the button but its not affecting the functionality

PR #1578, opened 2017-01-19
[fix bug 1331590] Change membership to pending for non open groups.

akatsoulas:
* PEP8 fixes.
* Improve code readability.
* Refactor notification system for membership changes.

### DIFF
--- a/mozillians/groups/models.py
+++ b/mozillians/groups/models.py
@@ -11,11 +11,12 @@ from mozillians.common.urlresolvers import reverse
 from mozillians.common.utils import absolutify
 from mozillians.groups.managers import GroupBaseManager, GroupQuerySet
 from mozillians.groups.templatetags.helpers import slugify
-from mozillians.groups.tasks import email_membership_change, member_removed_email
+from mozillians.groups.tasks import email_membership_change
 from mozillians.users.tasks import unsubscribe_from_basket_task, subscribe_user_to_basket
 
 
 class GroupBase(models.Model):
+    """Base class for groups in Mozillians."""
     name = models.CharField(db_index=True, max_length=50,
                             unique=True, verbose_name=_lazy(u'Name'))
     url = models.SlugField(blank=True)
@@ -50,6 +51,8 @@ class GroupBase(models.Model):
         return results
 
     def save(self, *args, **kwargs):
+        """Override save method."""
+
         self.name = self.name.lower()
         super(GroupBase, self).save()
         if not self.url:
@@ -61,13 +64,15 @@ class GroupBase(models.Model):
         return self.name
 
     def merge_groups(self, group_list):
+        """Merge two groups."""
         for group in group_list:
-            map(lambda x: self.add_member(x),
-                group.members.all())
+            map(lambda x: self.add_member(x), group.members.all())
             group.aliases.update(alias=self)
             group.delete()
 
     def user_can_leave(self, userprofile):
+        """Checks if a member of a group can leave."""
+
         curators = self.curators.all()
         return (
             # some groups don't allow leaving
@@ -76,10 +81,11 @@ class GroupBase(models.Model):
             (curators.count() > 1 or userprofile not in curators) and
             # only makes sense to leave a group they belong to (at least pending)
             (self.has_member(userprofile=userprofile) or
-                self.has_pending_member(userprofile=userprofile))
+             self.has_pending_member(userprofile=userprofile))
         )
 
     def user_can_join(self, userprofile):
+        """Checks if a user can join a group."""
         return (
             # Must be vouched
             userprofile.is_vouched and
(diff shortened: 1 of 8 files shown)


akatsoulas:
@johngian r?

mozillians/groups/models.py
```diff
@@ -285,28 +315,41 @@ def add_member(self, userprofile, status=GroupMembership.MEMBER):
                                                    [settings.BASKET_NDA_NEWSLETTER])
 
     def remove_member(self, userprofile, send_email=True):
+        """Change membership status for a group.
+
+        If user is a member of an open group, then the user is removed.
+
+        If a user is a member of a reviewed or closed group,
```
johngian: I think that this complicates stuff a lot. ``remove_member`` should do what the name says.

mozillians/groups/models.py
```diff
@@ -285,28 +315,41 @@ def add_member(self, userprofile, status=GroupMembership.MEMBER):
...
+        # If the group is of type Group.OPEN, delete membership
+        if self.accepting_new_members == Group.OPEN or old_status != GroupMembership.MEMBER:
+            # We have either an open group or the request to join a reviewed group is denied
+            # or the curator manually declined a user in a pending state.
+            membership.delete()
+            # delete the invitation to the group if exists
+            Invite.objects.filter(group=self, redeemer=userprofile).delete()
+
+        # Group is either of Group.REVIEWED or Group.CLOSED, downgrade membership to PENDING
+        else:
```
johngian: What happens if we want to actually remove the member.
FYI this breaks the ``remove member from group`` functionality (eg. when an admin deletes a user, the user doesn't get deleted but goes to a pending state).

mozillians/groups/models.py
akatsoulas: I believe that we should follow the same paradigm as with `add_member` method which changes/upgrades membership depending on the status (pending, pending terms, member). That said, I don't mind renaming this to something clearer about its role. Having another entry point to change the membership status, sounds more complicated to me

mozillians/groups/models.py
akatsoulas: Agreed, I will have another look for this.

mozillians/groups/models.py
johngian: Ok let's keep the same paradigm. We can add a ``status`` argument which by default is None. The default functionality is to remove the member. If status exists it change's members status.

mozillians/groups/models.py
```diff
@@ -284,29 +314,49 @@ def add_member(self, userprofile, status=GroupMembership.MEMBER):
...
             return
         old_status = membership.status
-        membership.delete()
+        new_status = None
+
+        # If the group is of type Group.OPEN, delete membership
+        # If no status is given, delete membership,
+        # If the current membership is PENDING*, delete membership
+        if (not status or self.accepting_new_members == Group.OPEN
+                or old_status != GroupMembership.MEMBER):
```
johngian: Why do we need ``old_status != GroupMembership.MEMBER`` check?

mozillians/groups/models.py
```diff
@@ -284,29 +314,49 @@ def add_member(self, userprofile, status=GroupMembership.MEMBER):
...
+            # We have either an open group or the request to join a reviewed group is denied
+            # or the curator manually declined a user in a pending state.
+            membership.delete()
+            # delete the invitation to the group if exists
+            Invite.objects.filter(group=self, redeemer=userprofile).delete()
+
+        # Group is either of Group.REVIEWED or Group.CLOSED, downgrade membership to PENDING
+        else:
+            membership.status = GroupMembership.PENDING
```
johngian: nit: You can rewrite this like:
```
membership.status = status or GroupMembership.PENDING
new_status = status or GroupMembership.PENDING
```

mozillians/groups/models.py
```diff
@@ -284,29 +314,49 @@ def add_member(self, userprofile, status=GroupMembership.MEMBER):
...
+            new_status = GroupMembership.PENDING
```
johngian: Why do we need to assign ``new_status`` ? We are assigning it the same value with ``membership.status`` which never changes until the end of the function.

mozillians/groups/models.py
akatsoulas: Because if a user is in a pending state (PENDING_TERMS or just PENDING) and there is a removal, then we need to further downgrade the user. This could happen in the edge case where a request is in a pending state for a period longer than the invalidation period of the group.

johngian: Yes but this would happen anyway because ``status=None``.

akatsoulas: Nope, because the invalidate_membership method just passes `status=GroupMembership.PENDING` if the group is closed or reviewed. It doesn't check the individual status of every user. This is a job for this method. So if we factor in this, we could have the following scenario:
- User JohnDoe is in a pending state for the closed group `foo`.
- The memberships of group `foo` start to expire.
- Everyone whose membership expires should go into PENDING state.
- JohnDoe would follow this paradigm when this user should be just dropped following the downgrade schema.

johngian:
Code looks OK. Let's test it on dev.

johngian:
Please fix failing tests.

PR #1585, opened 2017-01-26
Add a waffle flag for invalidation testing.

akatsoulas:


mozillians/groups/views.py
```diff
@@ -576,3 +578,30 @@ def send_invitation_email(request, invite_pk):
...
+
+@never_cache
+@waffle_flag('force_group_invalidation', '404')
+def force_group_invalidation(request, url, alias_model, template=''):
+    """View to help test different scenarios.
+
+    Forces an immediate invalidation in the case that a group is
+    set to expire its memberships.
+    """
+    group_alias = get_object_or_404(alias_model, url=url)
```
johngian: Why don't you use group.name directly?

mozillians/groups/views.py
```diff
@@ -576,3 +578,30 @@ def send_invitation_email(request, invite_pk):
...
+    group = group_alias.alias
+    is_curator = request.user.userprofile in group.curators.all()
```
johngian: nit: I would prefer ``group.curators.filter()``

mozillians/groups/views.py
akatsoulas: We need the alias model first in order to distinguish `groups` from `skills`.

akatsoulas:
@johngian pr updated

PR #1591, opened 2017-01-27
Explain in group email templates when a membership has expired.

akatsoulas:


akatsoulas:
@johngian thoughts? should we include this in our templates?

akatsoulas:
@johngian thoughts on this?

PR #1601, opened 2017-02-13
Allow testing nda renewal notifications.

johngian:


johngian:
@akatsoulas r?

mozillians/groups/tasks.py
```diff
@@ -161,16 +162,17 @@ def notify_membership_renewal():
                        .exclude(userprofile__id__in=curator_ids))
 
         # Filter memberships to be notified
-        last_update_days = group.invalidation_days - days_prior_invalidation
-        last_update = now() - timedelta(days=last_update_days)
+        if switch_is_active('test_group_invalidation_email'):
```
akatsoulas: Shouldn't this be `if not`?

mozillians/groups/views.py
```diff
@@ -608,3 +609,10 @@ def force_group_invalidation(request, url, alias_model, template=''):
         raise Http404
 
     return redirect(reverse('groups:show_group', args=[group.url]))
+
+
+@never_cache
+@waffle_switch('test_membership_renewal_notification')
+def membership_renewal_notification(request):
```
akatsoulas: Let's change the view to `test_membership_renewal_notification` or add a docstring explaining why is this view here and what it does.

mozillians/groups/tasks.py
akatsoulas: Let's add a comment to this if clause.

PR #1606, opened 2017-02-14
Flag profiles ready to expire eligible for an early renewal.

akatsoulas:
* minor changes to help readabiity.

### DIFF
--- a/mozillians/groups/models.py
+++ b/mozillians/groups/models.py
@@ -1,6 +1,7 @@
 from django.core.exceptions import ValidationError
 from django.conf import settings
 from django.db import models
+from django.db.models import Q
 from django.utils.timezone import now
 from django.utils.translation import ugettext as _
 from django.utils.translation import ugettext_lazy as _lazy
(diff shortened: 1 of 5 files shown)


johngian:
Tested it locally:

* It doesn't allow you to renew membership for members
* There is no way for curators to filter only pending renewals

akatsoulas:
@johngian pr updated. The membership of existing members can be renewed. As for the second comment, let's tackle it in another PR since it seems to me a bit out of scope.

mozillians/groups/models.py
```diff
@@ -305,6 +309,7 @@ def add_member(self, userprofile, status=GroupMembership.MEMBER):
                         (GroupMembership.PENDING_TERMS, GroupMembership.MEMBER)]
             if (old_status, status) in statuses:
                 # Status changed
+                # let's remove the needs renewal flag
```
johngian: Please move this comment to line 295

johngian:
@akatsoulas I pushed a patch here to allow filtering renewals please review and apply:

https://github.com/johngian/mozillians/commit/b3aedc03dbda57abab95365feda7da7cc612b54f

PR #1613, opened 2017-02-15
Send emails only to inviters.

johngian:


mozillians/groups/tasks.py
```diff
@@ -201,10 +201,14 @@ def notify_membership_renewal():
             format_args = [membership.group.name, membership.userprofile.full_name]
             subject = _(subject_msg.format(*format_args))
 
-            for curator in group.curators.all():
```
akatsoulas: With this patch, we will end up sending notifications only to the curators of a `CLOSED` group, completely ignoring the curators of a `REVIEWED` group. If we want to target the curator that invited a user in a closed group then we should make sure that we are handling somehow the curators of the reviewed groups too.

mozillians/groups/tasks.py
```diff
@@ -201,10 +201,14 @@ def notify_membership_renewal():
...
-                ctx['curator_full_name'] = curator.full_name
-                message = curator_template.render(ctx)
-                send_mail(subject, message, settings.FROM_NOREPLY, [curator.email])
+            invites = Invite.objects.filter(group=group, redeemer=membership.userprofile)
```
akatsoulas: This should be one invite so probably `invite` var makes more sense. That said, I would go for `Invite.objects.get` with a `try except`

akatsoulas:
After a discussion with @johngian and @hmitsch this is out of scope. Closing.

PR #1617, opened 2017-02-20
Add django-cities-light support

johngian:


johngian:
@akatsoulas PR updated with an ``update.py`` command to populate geonames

mozillians/users/api/v2.py
```diff
@@ -108,9 +108,13 @@ def __init__(self, *args, **kwargs):
         # If we don't define a custom transform method and if the
         # field has a privacy setting, set the transform privacy
         # wrapper.
+
         for field in self.fields.keys():
             method_name = 'transform_{0}'.format(field)
 
+            if field in ['country', 'region', 'city']:
```
akatsoulas: Why is this needed?

johngian: Because we already define ``country``, ``region`` and ``city`` as aliases to ``geo_country``, ``geo_region``, ``geo_city``and it tries to find the privacy fields for that fields instead.

akatsoulas: Please add a comment explaining what this is doing.

scripts/update/update.py
```diff
@@ -49,6 +49,7 @@ def update_assets(ctx):
         ctx.local('LANG=en_US.UTF-8 python manage.py collectstatic --noinput --no-default-ignore -i .git')  # noqa
         ctx.local('LANG=en_US.UTF-8 python manage.py compress --engine jinja2')
         ctx.local('LANG=en_US.UTF-8 python manage.py update_product_details')
+        ctx.local('LANG=en_US.UTF-8 python manage.py cities_light_fixtures load --base-url=lib/django_cities_light/fixtures/')  # noqa
```
johngian: This needs to happen after the migrations

PR #1619, opened 2017-02-22
Try to send the inviter email first else fallback to all curators.

johngian:


mozillians/groups/tasks.py
```diff
@@ -208,7 +208,9 @@ def notify_membership_renewal():
             format_args = [membership.group.name, membership.userprofile.full_name]
             subject = _(subject_msg.format(*format_args))
 
-            for curator in group.curators.all():
+            curators = [ctx['inviter']] if ctx['inviter'] else group.curators.all()
```
akatsoulas: This can be a bit difficult to read because it looks like a list comprehension. Maybe something like:

```python
curators = group.curators.all()
if ctx['inviter']:
    curators = [ctx['inviter']]
```
is more readable

akatsoulas: also please add a comment to explain why we are changing the recipients based on the type of the group.

mozillians/groups/tests/test_tasks.py
```diff
@@ -494,6 +494,40 @@ def test_send_renewal_notification_curators_email(self, mock_now, mock_send_mail
...
+    @patch('mozillians.groups.tasks.send_mail')
+    @patch('mozillians.groups.tasks.now')
+    def test_send_renewal_notification_inviters_email(self, mock_now, mock_send_mail):
+        """Test renewal notification functionality for curators"""
+        curator1 = UserFactory.create(email='foo@example.com')
+        curator2 = UserFactory.create(email='foobar@example.com')
+        curator3 = UserFactory.create(email='bar@example.com')
+        member = UserFactory.create(userprofile={'full_name': 'Example Name'})
+        group = GroupFactory.create(name='foobar', invalidation_days=365,
+                                    accepting_new_members=Group.REVIEWED)
```
akatsoulas: A `Reviewed` group does not have invites. That said, shall we refactor the task to distinguish the functionality between group types instead of using `get_object_or_none`? The end result will be the same but it will be more clear where the difference in functionality between groups lies. 

Personally I would prefer a more clear distinction between groups but it's up to you. Let's make sure that we have proper documentation in the code.

PR #1625, opened 2017-02-28
Upgrade Celery to version 4.0.2

comzeradd:


akatsoulas:
This PR does not handle the periodic tasks. We need to migrate all the periodic tasks to the new API
http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html

mozillians/settings/base.py
```diff
@@ -483,23 +482,17 @@ def _allowed_hosts():
 DEFAULT_AVATAR_URL = urljoin(MEDIA_URL, DEFAULT_AVATAR)
 DEFAULT_AVATAR_PATH = os.path.join(MEDIA_ROOT, DEFAULT_AVATAR)
 
-# Celery configuration
-import djcelery  # noqa
-djcelery.setup_loader()
-CELERYBEAT_SCHEDULER = 'djcelery.schedulers.DatabaseScheduler'
```
akatsoulas: This should not be part of this PR

mozillians/celery.py
```diff
@@ -11,5 +11,16 @@
 
 app = Celery('mozillians')
 
-app.config_from_object('django.conf:settings')
+app.config_from_object('django.conf:settings', namespace='CELERY')
 app.autodiscover_tasks(lambda: settings.INSTALLED_APPS)
+
+app.conf.beat_schedule = {
```
akatsoulas: Why don't you use the `add_periodic_task` function here?

mozillians/settings/base.py
comzeradd: I removed it for now and we can rebase later, but with `django-celery` installed tests will fail.

mozillians/celery.py
comzeradd: To avoid importing the tasks here.

akatsoulas: I am not sure I understand this. The `add_periodic_task` will do exactly the same thing behind the scenes. It is just a more pythonic way to register the periodic tasks.

http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html#entries

mozillians/celery.py
```diff
@@ -11,5 +11,16 @@
...
+    'invalidate-group-membership': {
+        'task': 'mozillians.groups.tasks.invalidate_group_membership',
+        'schedule': 60
```
akatsoulas: The task is scheduled to run every 24 hours.

mozillians/celery.py
```diff
@@ -11,5 +11,16 @@
...
+    },
+    'notify-membership-renewal': {
+        'task': 'mozillians.groups.tasks.notify_membership_renewal',
+        'schedule': 60
```
akatsoulas: same here

mozillians/settings/base.py
akatsoulas: cool, then let's add it and when the other pr is merged a rebase with the master will do the trick

mozillians/users/tasks.py
```diff
@@ -189,7 +189,7 @@ def subscribe_user_to_basket(instance_id, newsletters=[]):
 
     lookup_subtask = lookup_user_task.subtask((instance.user.email,))
     subscribe_subtask = subscribe_user_task.subtask((instance.user.email, newsletters,))
-    chain(lookup_subtask | subscribe_subtask)()
+    chain(lookup_subtask, subscribe_subtask)()
```
akatsoulas: According to celery's documentation, pipes should work fine. Is there another reason for changing this?

comzeradd: True. Both will work. I did for simplicity/readability reasons. According to the [documentation](http://docs.celeryproject.org/en/latest/userguide/canvas.html#chains) chains should either be in the form: `chain(task, task)` or just: `(task | task)`. So no need to use pipes when you use the chain keyword.

mozillians/celery.py
akatsoulas: Let's try to use the `add_periodic_task` and keep the `beat_schedule` as a backup option.

akatsoulas:
@comzeradd let's rebase this to master.

comzeradd:
Rebased

mozillians/celery.py
```diff
@@ -11,5 +11,16 @@
...
-app.config_from_object('django.conf:settings')
+app.config_from_object('django.conf:settings', namespace='CELERY')
 app.autodiscover_tasks(lambda: settings.INSTALLED_APPS)
+
+
+@app.on_after_finalize.connect
+def setup_periodic_tasks(sender, **kwargs):
+    from mozillians.groups.tasks import notify_membership_renewal, invalidate_group_membership
+
+    sender.add_periodic_task(60 * 60 * 24, invalidate_group_membership.s(),
```
akatsoulas: Let's move the `60*60*24` to a var on the top of the file, something like `RUN_DAILY`

comzeradd:
PR updated

PR #1632, opened 2017-03-02
Add datamigration from mozillians.geo data to cities_light.

johngian:


johngian:
After the migration I got this stats:
```
In [8]: UserProfile.objects.filter(country__isnull=True, geo_country__isnull=False).count()
Out[8]: 30

In [9]: UserProfile.objects.filter(city__isnull=True, geo_city__isnull=False).count()
Out[9]: 396

In [10]: UserProfile.objects.filter(region__isnull=True, geo_region__isnull=False).count()
Out[10]: 2237

In [11]: UserProfile.objects.all().count()
Out[11]: 12653
```

mozillians/users/admin.py
```diff
@@ -272,6 +272,69 @@ def queryset(self, request, queryset):
         return queryset
 
 
+class MissingCountry(SimpleListFilter):
```
akatsoulas: Please drop anything related to the admin from this PR. Let's keep the data migration on one PR and the filters in another.

PR #1642, opened 2017-03-16
Do not allow object creation if skill exists.

johngian:


johngian:
@akatsoulas r?

mozillians/groups/views.py
```diff
@@ -501,7 +501,12 @@ class SkillsAutocomplete(autocomplete.Select2QuerySetView):
 
     def has_add_permission(self, request):
         """Return True if the user has the permission to add a model."""
-        if not request.user.is_authenticated():
+
+        obj_exists = False
+        if self.q:
+            obj_exists = Skill.objects.filter(name=self.q).exists()
```
akatsoulas: `create_object` looks like a better candidate for this specific functionality. You can check if the object exists there. If it does return it, otherwise create it. Thoughts?

PR #1650, opened 2017-03-19
Re-add location data in profile.

akatsoulas:
* Use django cities to populate drop-down menus.
* Use django autocomplete light for autocompletion.
* Filter options based on previous selection.


mozillians/jinja2/phonebook/edit_profile.html
```diff
@@ -35,11 +35,9 @@
               <a href="#profile" aria-controls="description" role="tab" data-toggle="tab">{{ _('Profile') }}</a>
             </li>
 
-            {% if waffle.flag('location_is_active') %}
               <li role="presentation" id="mylocation-tab">
```
johngian: Please fix identation.

mozillians/jinja2/phonebook/includes/profile_edit_location.html
```diff
@@ -13,6 +14,45 @@
         {% endif %}
 
         <fieldset>
+          {% for error in location_form.errors['location'] %}
+            <span class="error-message">{{ error }}</span>
+          {% endfor %}
+        </fieldset>
+
+        <fieldset id="country">
```
johngian: Why don't we use one fieldset for all the location related fields ? From MDN:
```
The HTML <fieldset> element is used to group several controls as well as labels (<label>) within a web form.
```

mozillians/static/mozillians/js/profile_edit.js
```diff
@@ -141,6 +141,31 @@ $(function() {
         });
     });
 
+    // Django autcomplete light - django cities light
+    (function() {
+        // If there are no location data, hide the fields
+        if ( !$(':input[name=country]').val() ){
```
johngian: Why do we need this? FYI we have legacy accounts with no location data. Also users might have created accounts while location data editing was flagged out. This means that they are never going to be able to fill this form.

mozillians/static/mozillians/js/profile_edit.js
```diff
@@ -141,6 +141,31 @@ $(function() {
...
+            $('#city').hide();
+            $('#region').hide();
+        }
+    })();
+
+    // Hide region and city dropdown on document load
```
johngian: Why don't we leave the fields as they are by default. Show/hide seems over-complicated and I don't see the point of having it.

mozillians/users/urls.py
```diff
@@ -16,4 +19,10 @@
         name='voucher-autocomplete'),
     url('vouched-autocomplete/$', login_required(VouchedAutocomplete.as_view()),
         name='vouched-autocomplete'),
+    url('country-autocomplete/$', login_required(CountryAutocomplete.as_view(model=Country)),
```
johngian: Can't we pass ``model==Country`` as ``class.Meta``?

mozillians/users/views.py
```diff
@@ -86,3 +87,51 @@ def get_queryset(self):
...
+
+    def get_queryset(self):
+        """Country queryset from cities_light."""
+
+        if not self.request.user.is_authenticated():
+            return Country.objects.none()
+        qs = Country.objects.all()
+
+        if self.q:
+            qs = qs.filter(name__istartswith=self.q)
```
johngian: Maybe it makes sense to add ``icontains`` for countries like ``Rebublic of X`` and user might start typing ``X``.

mozillians/users/views.py
```diff
@@ -86,3 +87,51 @@ def get_queryset(self):
...
+        """Region queryset from cities_light."""
+
+        country_id = self.forwarded.get('country')
+        if not self.request.user.is_authenticated() or not country_id:
+            return Region.objects.none()
+
+        country = Country.objects.get(id=country_id)
+        qs = Region.objects.filter(country=country)
+        if self.q:
+            qs = qs.filter(name__istartswith=self.q)
```
johngian: Same here eg. ``City of X``.

mozillians/users/views.py
```diff
@@ -86,3 +87,51 @@ def get_queryset(self):
...
+        region_id = self.forwarded.get('region')
+        if not self.request.user.is_authenticated() or not region_id:
+            return City.objects.none()
+
+        region = Region.objects.get(id=region_id)
+
+        qs = City.objects.filter(region=region, country=region.country)
+
+        if self.q:
+            qs = qs.filter(name__istartswith=self.q)
```
johngian: Same here.

mozillians/users/views.py
```diff
@@ -86,3 +87,51 @@ def get_queryset(self):
...
+    def get_queryset(self):
+        """City queryset from cities_light."""
+
+        region_id = self.forwarded.get('region')
+        if not self.request.user.is_authenticated() or not region_id:
+            return City.objects.none()
+
+        region = Region.objects.get(id=region_id)
+
+        qs = City.objects.filter(region=region, country=region.country)
```
johngian: From django-cities-light model:
https://github.com/yourlabs/django-cities-light/blob/master/cities_light/abstract_models.py#L148

johngian: City might have a ``null` ``Region`` or ``Country``. I suggest we try first this query and if it fails why try either ``region=region`` or ``country=country`` filters separately.

mozillians/jinja2/phonebook/includes/profile_edit_location.html
johngian: Feel free to omit since it had the same structure previously, but other than that seems better to have one fieldset with all location data.

johngian:
Please make sure fields are aligned.

johngian:
I suggest to add autocompletion in timezone too so we keep the same paradigm across fields.

PR #1651, opened 2017-03-20
Ensure CELERY_LOADER var is not set.

johngian:


johngian:
TLDR:

Previous commit that removed this line didn't remove variable from deployment environment. This should fix the issue.

PR #1655, opened 2017-03-21
Revert celery setup to version 3.1.24

johngian:


johngian:
@comzeradd can you take a look?

johngian:
I prefer to have different revert commits that reference the initial commit instead of a huge squash that's hard to track.

PR #1660, opened 2017-03-22
Filter cities by country in autocomplete, if region is not supplied.

akatsoulas:


mozillians/users/views.py
```diff
@@ -129,13 +129,18 @@ def get_queryset(self):
...
         if not self.request.user.is_authenticated():
             return City.objects.none()
 
         qs = City.objects.all()
         if region_id:
             region = Region.objects.get(id=region_id)
-            qs = City.objects.filter(region=region, country=region.country)
+            qs = qs.filter(region=region, country=region.country)
+
+        if country_id:
```
johngian: Shouldn't this be before ``if region_id`` ?

PR #1662, opened 2017-03-22
Fix squashed text when searching for location fields.

johngian:


johngian:
@akatsoulas r?

johngian:
@akatsoulas r?

PR #1666, opened 2017-03-22
Fix tracebacks in autocomplete.

akatsoulas:
* Raise the length of the name to 100 chars.
* Do not create a skill if already exists.


mozillians/groups/views.py
```diff
@@ -514,6 +514,9 @@ def get_create_option(self, context, q):
     def create_object(self, text):
         """Override create object method."""
 
+        # If the skill exists, just return it
+        if Skill.objects.filter(name=text).exists():
```
johngian: This can be done using a single SQL query

PR #1674, opened 2017-03-27
Build mozillians language lists

comzeradd:


mozillians/phonebook/tests/test_views/test_views_list.py
```diff
@@ -189,8 +190,69 @@ def test_list_mozillians_in_location_invalid_country(self):
...
+        eq_(response.context['people'].number, 1)
+
+    @patch('mozillians.groups.views.settings.ITEMS_PER_PAGE', 1)
+    def test_list_mozillians_by_language_vouched_empty_page(self):
+        user1 = UserFactory.create()
+        language = LanguageFactory.create(userprofile=user1.userprofile)
+        language.code = 'el'
+        language.save()
+        user = UserFactory.create()
+        url = reverse('phonebook:list_language', kwargs={'language': 'russian'})
```
johngian: Shouldn't ``russian`` be ``ru``?

mozillians/phonebook/views.py
```diff
@@ -540,14 +541,20 @@ def delete_apikey(request, api_pk):
     return redirect('phonebook:apikeys')
 
 
-def list_mozillians_in_location(request, country, region=None, city=None):
-    queryset = UserProfile.objects.vouched().filter(country__name__iexact=country)
-    show_pagination = False
+def list_filtered_mozillians(request, country=None, region=None, city=None, language=None):
+    if country:
```
johngian: Although this works fine. Is there anyway to make it more futureproof so we can be able to add more filters in the future? (i don't like the idea of adding another if/elif for the next filter parameter)

comzeradd: I can't think of an easy way to achieve that. We have a complexity element here, the county filter has subfilters so a minimum check is required anyway.

akatsoulas: One idea could be to make this a dynamic method. Specifically, you could pass an object (country, region, city, language) and from that object you can access the class to perform the queries. eg `obj.__class__.objects.filter`.
FYI, I haven't tested this and it's just an idea.

akatsoulas:
Is there an update on this?

comzeradd:
Since the search refactor we no longer have a location filter view. I refactored this to utilize haystack search for language lists too and deleted the unused template.

@akatsoulas @johngian do you think I should squash everything down to one commit?

comzeradd:
Commits squashed

comzeradd:
@johngian can you have another look at this?

PR #1680, opened 2017-04-03
[fix bug 1346959] Add source_url to basket subscribe task.

akatsoulas:


mozillians/users/tasks.py
```diff
@@ -32,6 +32,7 @@
 
 INCOMPLETE_ACC_MAX_DAYS = 7
 MOZILLIANS_NEWSLETTERS = [BASKET_NDA_NEWSLETTER, BASKET_VOUCHED_NEWSLETTER]
+MOZILLIANS_URL = 'https://mozillians.org'
```
johngian: Let's use ``SITE_URL`` instead since we have multiple envs doing basket calls.

PR #1685, opened 2017-04-06
Move user signals to a separate file.

akatsoulas:
* Load signals when the app is ready.


johngian:
After #1684 it looks like signals are getting triggered. I suggest we close this one.

akatsoulas:
Let's see what's wrong with the signals. After that we could separate the signals from the models for a more dry/cleaner approach. I am adding a DNM flag for now

PR #1687, opened 2017-04-06
Revert "Log autovouching flow on sentry."

johngian:
Reverts mozilla/mozillians#1684


johngian:
Let's remove the debug statements from the codebase.

PR #1688, opened 2017-04-07
Do not send pending notification to CLOSED groups.

akatsoulas:


mozillians/groups/tasks.py
```diff
@@ -68,9 +69,10 @@ def send_pending_membership_emails():
                 'count': count,
             })
 
-            send_mail(subject, body, settings.FROM_NOREPLY,
-                      [profile.user.email for profile in group.curators.all()],
-                      fail_silently=False)
+            if curators:
```
johngian: this check could be moved to line 52.

PR #1436, opened 2016-04-13
[Fix bug 1201027] Simplify Profile status UI

comzeradd:

PR #1695, opened 2017-04-22
[fix bug 1195698] Layout preview of profile doesn't look in the right order

tsmrachel:
https://bugzilla.mozilla.org/show_bug.cgi?id=1195698


tsmrachel:
notvalid : fixed with #1436

PR #1696, opened 2017-04-24
Add python decouple

flamingspaz:
Not tested yet but open for comments


mozillians/settings/env-dist
```diff
@@ -0,0 +1,76 @@
+# See base_env_vars for details about these settings
+DEV = False
```
akatsoulas: There is no need to include configuration variables that already have a default value. You can keep this file just to the minimum amount of vars needed in order to have a local instance running.

mozillians/settings/base_env_vars.py
```diff
@@ -0,0 +1,567 @@
...
+        'landing': {'level': logging.INFO},
+        'phonebook': {'level': logging.INFO},
+    },
+}
+
+# Repository directory
+ROOT = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+
+# path() bases things off of ROOT
+path = lambda *a: os.path.abspath(os.path.join(ROOT, *a))  # noqa
```
akatsoulas: You can remove this completely and use `Unipath`. There is an example on how to do so in the documentation.
eg https://pypi.python.org/pypi/python-decouple

requirements/prod.txt
```diff
@@ -335,3 +335,5 @@ progressbar2==3.12.0 \
 python-utils==2.0.1 \
     --hash=sha256:985f44edf24918d87531c339f8b126ce2d303cbbc9a4c7fc4dc81ac0726079ff \
     --hash=sha256:670f9b992ebfc005329fd9740408f2a704e595d64c8f8da83f7977618dba4acc
+python-decouple==3.0 \
```
akatsoulas: Since we are using hashes, we need to manually install any other dependencies. eg Unipath and dj-database-url

flamingspaz: Added those two, there doesn't seem to be any other deps for decouple

mozillians/settings/base_env_vars.py
flamingspaz: I think I got the path right here, can you verify it?

mozillians/settings/base_env_vars.py
```diff
@@ -0,0 +1,569 @@
...
+SYSLOG_TAG = config('SYSLOG_TAG', default="http_app_mozillians")
+LOGGING = {
+    'loggers': {
+        'landing': {'level': logging.INFO},
+        'phonebook': {'level': logging.INFO},
+    },
+}
+
+# Repository directory
+ROOT = Path(__file__).parent.parent
```
akatsoulas: Looks good!

flamingspaz:
So I think this should be working now, I get an error with compression but I think that might be my env. Could someone test?

I've also set `ALLOWED_HOSTS` to `DOMAIN` rather than stripping down `SITE_URL` (which is built from `DOMAIN` anyway)

mozillians/settings/base_env_vars.py
```diff
@@ -0,0 +1,562 @@
...
+    }
+}
+
+
+# Tells the product_details module where to find our local JSON files.
+# This ultimately controls how LANGUAGES are constructed.
+PROD_DETAILS_DIR = Path(config('PROD_DETAILS_DIR', default='lib/product_details_json')).resolve()
+
+# Accepted locales
+LANGUAGE_CODE = config('LANGUAGE_CODE', default='en-US')
```
akatsoulas: This is already defined in line 60

mozillians/settings/base_env_vars.py
```diff
@@ -0,0 +1,562 @@
...
+    from product_details import product_details
+
+    return [(lang.lower(), product_details.languages[lang]['native'])
+            for lang in get_langs() if lang in product_details.languages]
+
+
+LANGUAGES = lazy(lazy_langs, list)()
+
+# For absolute urls
+DOMAIN = config('DOMAIN', default='localhost')
```
akatsoulas: I would go with `mozillians.org` as the default value and add `localhost` in the env-dist as a default.

mozillians/settings/base_env_vars.py
```diff
@@ -0,0 +1,562 @@
...
+def COMPRESS_JINJA2_GET_ENVIRONMENT():
+    from django.template import engines
+    return engines['jinja2'].env
+
+
+# Absolute path to the directory that holds media.
+MEDIA_ROOT = Path('media').resolve()
+
+# URL that handles the media served from MEDIA_ROOT. Make sure to use a
+MEDIA_URL = '/media/'
```
akatsoulas: MEDIA_URL = config('MEDIA_URL', default='/media/')

mozillians/settings/base_env_vars.py
```diff
@@ -0,0 +1,562 @@
...
+
+# Absolute path to the directory static files should be collected to.
+STATIC_ROOT = Path('static').resolve()
+
+# URL prefix for static files.
+STATIC_URL = '/static/'
```
akatsoulas: Similar to MEDIA_URL.

mozillians/settings/base_env_vars.py
```diff
@@ -0,0 +1,562 @@
...
+BASE_PASSWORD_HASHERS = (
+    'django_sha2.hashers.BcryptHMACCombinedPasswordVerifier',
+    'django_sha2.hashers.SHA512PasswordHasher',
+    'django_sha2.hashers.SHA256PasswordHasher',
+    'django.contrib.auth.hashers.SHA1PasswordHasher',
+    'django.contrib.auth.hashers.MD5PasswordHasher',
+    'django.contrib.auth.hashers.UnsaltedMD5PasswordHasher',
+)
+
+PWD_ALGORITHM = 'bcrypt'
```
akatsoulas: Let's add a config here too with a default value of `bcrypt`.

mozillians/settings/env-dist
```diff
@@ -0,0 +1,24 @@
...
+DOMAIN = 'localhost'
+PROTOCOL = 'https://'
+PORT = 443
+ES_DISABLED = True
+ES_URLS = 'http://127.0.0.1:9200'
+ES_INDEXING_TIMEOUT = 10
+CELERY_RESULT_BACKEND = 'amqp'
+CELERY_ACCEPT_CONTENT = 'pickle'
+CELERYD_TASK_SOFT_TIME_LIMIT = 150
+SECRET_KEY = ''
```
akatsoulas: Please add a non empty default value. Extra points for being creative!

akatsoulas:
Let's also add `.env` in gitignore.

mozillians/settings/env-dist
akatsoulas: Actually let's assign the default value of `True` here since this is going to be used for local dev.

mozillians/settings/env-dist
```diff
@@ -0,0 +1,24 @@
...
+USE_TZ = True
+ITEMS_PER_PAGE = 24
+COMPRESS_OFFLINE = True
+COMPRESS_ENABLED = True
+OIDC_RP_CLIENT_SECRET_ENCODED = True
+OIDC_STORE_ACCESS_TOKEN = True
+OIDC_OP_DOMAIN = 'auth.mozilla.auth0.com'
+OIDC_RP_CLIENT_ID = ''
+OIDC_RP_CLIENT_SECRET = ''
```
akatsoulas: Let's also add the following in this file:
COMPRESS_ENABLED = False
COMPRESS_OFFLINE = False
SESSION_COOKIE_SECURE=False
SITE_URL = http://127.0.0.1:8000
ALLOWED_HOSTS = 127.0.0.1

akatsoulas:
Fantastic job!! 🚀

mozillians/settings/base_env_vars.py
```diff
@@ -0,0 +1,560 @@
...
+            for lang in get_langs() if lang in product_details.languages]
+
+
+LANGUAGES = lazy(lazy_langs, list)()
+
+# For absolute urls
+DOMAIN = config('DOMAIN', default='mozillians.org')
+PROTOCOL = config('PROTOCOL', default='https://')
+PORT = config('PORT', default=443, cast=int)
+SITE_URL = config('SITE_URL', default='https://mozillians.org'
```
akatsoulas: Missing parentheses at the end of the line.

PR #1697, opened 2017-04-27
Make mozillians dockerfiles prod-ready.

johngian:
* Create new Dockerfiles for dev/prod
* Update docker-compose to use new Dockerfiles
* Create link for wsgi.py
* Install gunicorn dependency


docker/dev
```diff
@@ -0,0 +1,15 @@
+FROM python:2.7-slim
+
+EXPOSE 8000
+WORKDIR /code
+CMD ["./bin/run-dev.sh"]
+
+RUN adduser --uid 431 --disabled-password --disabled-login --gecos 'webdev' --no-create-home webdev
```
akatsoulas: Since we are running it as root, we don't need this for the dev Dockerfile.

wsgi/playdoh.wsgi
```diff
@@ -20,7 +20,7 @@ os.environ.setdefault('CELERY_LOADER', 'django')
 wsgidir = os.path.dirname(__file__)
 site.addsitedir(os.path.abspath(os.path.join(wsgidir, '../')))
 
-from django.core.wsgi import get_wsgi_application
+from django.core.wsgi import get_wsgi_application  # noqa
```
akatsoulas: Out of curiosity, why do we need the `noqa` flag here?

docker-compose.yml
```diff
@@ -1,50 +1,57 @@
...
+  db:
+    image: mysql:5.5
+    command: mysqld --character-set-server=utf8 --collation-server=utf8_unicode_ci --max_allowed_packet=3M
+    environment:
+      - MYSQL_DATABASE=mozillians
+      - MYSQL_USER=mozillians
+      - MYSQL_PASSWORD=mozillians
+      - MYSQL_ROOT_PASSWORD=root
+  es:
+    image: elasticsearch:1.7
```
akatsoulas: I believe the official image is deprecated.
https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html

wsgi/playdoh.wsgi
johngian: Pep8 error because the import is not on top

docker-compose.yml
johngian: Let's use this for now and upgrade it after haystack search refactor.

PR #1703, opened 2017-05-15
 Compress HTML templates & add whitenoise

flamingspaz:


mozillians/settings/base_env_vars.py
```diff
@@ -148,7 +148,7 @@ def lazy_langs():
         'OPTIONS': {
             'debug': DEBUG,
             'app_dirname': 'jinja2',
-            'match_extension': None,
+            'match_extension': '.html',
```
flamingspaz: It's nice to have as a fallback in case someone forgets to specify in the command line

PR #1706, opened 2017-05-29
Update profile index on update/delete.

akatsoulas:


mozillians/users/signals.py
```diff
@@ -0,0 +1,91 @@
...
+def add_employee_vouch(sender, instance, **kwargs):
+    """Add a vouch if an alternate email address is a mozilla* address."""
+
+    if kwargs.get('raw') or not instance.type == ExternalAccount.TYPE_EMAIL:
+        return
+    instance.user.auto_vouch()
+
+
+# Django Haystack signals
+class ProfileSignalProcessor(BaseSignalProcessor):
```
akatsoulas: Haystack's documentation suggests that the signal processor should not be inside models.py, thus the separation.

akatsoulas:
@johngian r?

mozillians/users/signals.py
```diff
@@ -0,0 +1,91 @@
...
+from haystack.signals import BaseSignalProcessor
+from multidb.pinning import use_master
+
+from mozillians.users.models import ExternalAccount, UserProfile, Vouch
+from mozillians.users.tasks import subscribe_user_to_basket, unsubscribe_from_basket_task
+
+
+# Signal to create a UserProfile.
+@receiver(signals.post_save, sender=User,
+          dispatch_uid='create_user_profile_sig')
```
johngian: Single line please

mozillians/users/signals.py
```diff
@@ -0,0 +1,91 @@
...
+    if kwargs.get('raw') or not instance.type == ExternalAccount.TYPE_EMAIL:
+        return
+    instance.user.auto_vouch()
+
+
+# Django Haystack signals
+class ProfileSignalProcessor(BaseSignalProcessor):
+
+    def setup(self):
+        signals.post_save.connect(self.handle_save, sender=UserProfile)
```
johngian: Should we also handle cases where obj is not saved but updated?

akatsoulas: We don't use the `update` method anywhere in the `users` app. Moreover, in order to handle this case we should be a bit creative since the `update` method does not emit a signal

PR #1708, opened 2017-05-30
Enable search for groups along with the user profiles.

akatsoulas:


mozillians/jinja2/search/indexes/groups/group_text.txt
```diff
@@ -1,3 +1,4 @@
 {{ object.name }}
 {{ object.description }}
 {{ object.wiki }}
+{{ object.visible }}
```
johngian: I am not sure if we need a boolean field that translates to ``True`` or ``False`` in a free text search.

PR #1713, opened 2017-06-01
Limit search based on the selection of advanced fields.

akatsoulas:


akatsoulas:
@johngian r?

mozillians/phonebook/forms.py
```diff
@@ -464,6 +465,16 @@ def search(self):
         """Search on the ES index the query sting provided by the user."""
 
         search_term = self.cleaned_data['q']
+        search_in_groups = False
+        search_in_profiles = False
+        search_all = True
+
+        for model in self.get_models():
+            search_all = False
```
johngian: Isn't this always assigned as ``False`` this way?

akatsoulas: No, the `self.get_models()` is only populated if any of the checkboxes in the advanced fields is checked, otherwise is an empty list.

PR #1715, opened 2017-06-02
Fix anonymous search in haystack.

akatsoulas:


mozillians/phonebook/forms.py
```diff
@@ -485,14 +493,13 @@ def search(self):
         for p_field in privacy_indexed_fields:
             # this is the field that we are going to query
             q_field = p_field.split('_', 1)[1]
-            if hasattr(profile, q_field):
```
johngian: Why are you removing this check? We need to make sure that object has the privacy field as attribute.

akatsoulas: The privacy fields are derived from the privacy_indexed_fields. These fields are coming from the index which means that all the profiles in mozillians.org are going to have one since we are excluding the non complete profiles. Do you believe there is a case where the privacy attribute can be missing?

akatsoulas: Better said, the profiles in the index will always have these privacy fields since we are excluding from indexing non complete profiles.

PR #1716, opened 2017-06-06
Exclude groups from public search.

akatsoulas:


akatsoulas:
@johngian pr is updated. can you have another look?

mozillians/phonebook/forms.py
```diff
@@ -493,31 +475,42 @@ def search(self):
...
-                # (lower number, means greater permission level)
-                q_args = {
-                    q_field: search_term,
-                    '{0}__gte'.format(p_field): privacy_level
-                }
-                query.add(SQ(**q_args), SQ.OR)
+        all_indexed_fields = UserProfileIndex.fields.keys()
+        privacy_indexed_fields = [field for field in all_indexed_fields
+                                  if field.startswith('privacy_')]
+        # Every profile object in mozillians.org has privacy sjettings.
```
johngian: *sjettings

mozillians/phonebook/forms.py
```diff
@@ -493,31 +475,42 @@ def search(self):
...
+                q_field: search_term,
+                '{0}__gte'.format(p_field): privacy_level
+            }
+            query.add(SQ(**q_args), SQ.OR)
 
         # Group Search
-        # We need to exclude non visible groups.
-        if search_all or search_in_groups:
-            query.add(SQ(**{'visible': True}), SQ.OR)
+        query.add(SQ(**{'visible': True}), SQ.OR)
```
johngian: What happens here if we only search on ``UserProfile`` index?

johngian: Please add a check if required to make sure we are also searching for ``Groups``.

johngian:
r+wc

mozillians/phonebook/forms.py
akatsoulas: Due to the `SQ.OR` the query will be ignored. Specifically, if we don't search in groups, the `sqs` will return results only from profiles which will not match this query and we won't have any problems due to the `OR` part. The only way to affect the search results will be if we add a `visible` flag in the profiles at some point.

PR #1719, opened 2017-06-07
Exclude pending members from member_count.

akatsoulas:


johngian:
@akatsoulas Is this PR still valid? Tests are failing

akatsoulas:
It is, I will spend some time updating them.

PR #1725, opened 2017-06-14
Add aws_host param for AWS signing

flamingspaz:


mozillians/settings/base_env_vars.py
```diff
@@ -579,6 +579,7 @@ def _lazy_haystack_setup():
         from elasticsearch import RequestsHttpConnection
 
         auth = AWSRequestsAuth(
+            aws_host=es_url,
```
johngian: I don't think that this is going to work with the full url. It only needs the domain.

mozillians/settings/base_env_vars.py
```diff
@@ -566,20 +558,24 @@ def _username_algo(email):
 
 
 # Django Haystack
+ES_DISABLED = config('ES_DISABLED', default=True)
+ES_HOST = config('ES_HOST',
```
johngian: Single line please.

mozillians/settings/base_env_vars.py
```diff
@@ -566,20 +558,24 @@ def _username_algo(email):
...
+                 default='127.0.0.1:9200')
+ES_PROTOCOL = config('ES_PORT', default='http://')
```
johngian: For consistency please use the same env var name

PR #1728, opened 2017-06-16
Enable django-axes to protect /admin.

johngian:


johngian:
@akatsoulas r?

Fixes https://github.com/mozilla/parsys/issues/175

mozillians/settings/base_env_vars.py
```diff
@@ -608,3 +609,6 @@ def _lazy_haystack_setup():
 
 HAYSTACK_CONNECTIONS = lazy(_lazy_haystack_setup, dict)()
 HAYSTACK_SIGNAL_PROCESSOR = 'mozillians.common.signals.SearchSignalProcessor'
+
+# Setup django-axes
+AXES_BEHIND_REVERSE_PROXY = True
```
akatsoulas: Is the proxy configured for this?

akatsoulas: Also, I would suggest to block IPs after a number of failed login attempts. Moreover, is there a way to hide admin behind ldap?

johngian: Proxy is configured + blocks after failing attempts by default.

I don't think there is a straightforward way to use admin for LDAP in AWS. It works only in SCL3. Also using LDAP is not suggested now with auth0.

PR #1734, opened 2017-06-19
Add cities_light loading to docker installation instructions

claudijd:
Was trying to get a docker instance up and running, and followed the docker installation instructions.  However, mozillians requires you to update the superuser profile you create in order to start using the app.  To do this, you need the country/region/city data loaded in the DB or it won't allow the user to advance through.  This update to the document is an attempt to make it so that others don't get hung up on this.


johngian:
Thanks for the PR @claudijd 
r+ :rocket:

PR #1737, opened 2017-06-19
Enable geo search from profiles.

akatsoulas:


mozillians/phonebook/forms.py
```diff
@@ -460,13 +460,30 @@ def __init__(self, *args, **kwargs):
...
+            location_query['privacy_country__gte'] = None
+        if self.region:
+            location_query['region'] = self.region
+            location_query['privacy_region__gte'] = None
+        if self.city:
+            location_query['city'] = self.city
+            location_query['privacy_city__gte'] = None
+
+        if not search_term:
+            search_term = '{0} {1} {2}'.format(self.country, self.region, self.city)
```
johngian: Instead of manually searching i would suggest to use ``all()`` and filter.

PR #1744, opened 2017-06-19
Update wording when displaying search results.

akatsoulas:


TonnesM:
https://github.com/mozilla/mozillians/commit/f4f00370f8e6d52d9df43916f5963e8ccac008cb
https://github.com/mozilla/mozillians/blob/master/mozillians/jinja2/phonebook/search.html#L113

The "related to" part seems to be hardcoded now and canot be localized, even when including it in l10n after "profile". Shouldn’t it be, like previously?

Also note that some locales (like nl) use double quotes for spoken/written text quotes from people only (i.e. literal quotes) and single ones in other cases - both curly when possible. Hence if the quote characters can be localized too like before, that would be nice.

akatsoulas:
@TonnesM thank you for catching this! PR #1758 adds reverts this error.

TonnesM:
@akatsoulas Thank you. I also noticed that PR was reverted in PR #1764, probably for a reason. Just curious, will such change land eventually?

akatsoulas:
It will, today :)
There was an syntax error which affected the server.

PR #1763, opened 2017-06-21
Upgrade mozilla-django-oidc to latest version.

johngian:


akatsoulas:
@johngian is this good to go?

PR #1768, opened 2017-06-21
Allow loading images from MEDIA_URL

flamingspaz:


mozillians/settings/base_env_vars.py
```diff
@@ -411,6 +411,7 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
     '*.google-analytics.com',
     '*.gravatar.com',
     '*.wp.com',
+    config('MEDIA_URL', default=''),
```
johngian: You can use MEDIA_URL directly

johngian: Although ``MEDIA_URL`` might have a path too. Let's hardcode it instead

PR #1784, opened 2017-06-30
[Fix bug 1376626] Remove extra space on Region/City list

comzeradd:


mozillians/jinja2/phonebook/search.html
```diff
@@ -69,7 +69,7 @@
       {% if country or region or city  %}
         <h2>{{ _('Mozillians in') }}
           {% for name in [region, city] %}
-            {% if name %}{{ name }} ,{% endif %}
+            {% if name %}{{ name }},{% endif %}
```
akatsoulas: If you want to control how this is rendered, you can also use Jinja's whitespace control
http://jinja.pocoo.org/docs/2.9/templates/#whitespace-control

comzeradd: Yes, but in this case we don't need to over-complicate it. We would never need a whitespace right before a comma.

akatsoulas: That's true but the extra space helps the readability of the code. I would go a step further and I would add an extra space before the `endif` statement too. Using the builtin functionality is not an over-complication. 

Since this is nit I am approving the PR.

PR #1790, opened 2017-07-07
Use two gunicorn workers per instance.

akatsoulas:


akatsoulas:
@flamingspaz thoughts on this?

PR #1792, opened 2017-07-14
Spam deletion

johngian:


mozillians/settings/base_env_vars.py
```diff
@@ -23,7 +23,8 @@
 DEV = config('DEV', default=False, cast=bool)
 DEBUG = config('DEBUG', default=False, cast=bool)
 
-ADMINS = ()
+ADMIN_ALIAS = config('ADMIN_ALIAS', default='mozillians-admins@mozilla.com')
```
akatsoulas: Why do you need the `ADMIN_ALIAS` var? You could set directly the address to the `ADMINS` var

mozillians/users/tasks.py
```diff
@@ -264,6 +265,52 @@ def check_spam_account(instance_id, **kwargs):
...
+@task
+def delete_spam_account():
+    """Task to automatically delete spam accounts"""
+
+    from mozillians.users.models import AbuseReport
+
+    reports = AbuseReport.objects.all()
+
+    # Manual reports deletion heuristic:
+    # * Delete all unvouched profiles
```
akatsoulas: super nit: I would remove the `*` from the comment. My only argument is that it looks a bit funny to me. You can of course completely ignore this comment

mozillians/users/tasks.py
```diff
@@ -264,6 +265,52 @@ def check_spam_account(instance_id, **kwargs):
...
+    query = {
+        'is_akismet': True,
+        'profile__is_vouched': False,
+        'profile__skills__isnull': True,
+    }
+    reports.filter(**query).delete()
+
+    # Email admins on manual reports for vouched profiles
+
+    query = {
```
akatsoulas: Since we have reports for this (w00t w00t), I would go for a `Q` filter here to add also the profiles marked as spam by akismet but have skills. This way we won't allow this specific category to fall through the cracks.

akatsoulas:
Also travis is a bit grumpy.

mozillians/settings/base_env_vars.py
johngian: There is no way to parse ``tuple`` in decouple and it gets way too complicated to implement the parser.

PR #1793, opened 2017-07-17
Improve ES index rebuilding

johngian:


johngian:
@akatsoulas r?

mozillians/users/tasks.py
```diff
@@ -320,4 +320,30 @@ def delete_spam_account():
...
+        'workers': 10,
+        'batchsize': 100,
+        'interactive': False,
+    }
+
+    # Rebuild index in ES_INDEX_HELPER
+    call_command('rebuild_index', **options)
+
+    # Copy documents from `helper` ES to `default` ES
+    from elasticsearch import helpers
```
akatsoulas: Please move those either to the top of the file or to the top of the task if they are causing circular dependencies.

mozillians/users/tasks.py
```diff
@@ -320,4 +320,30 @@ def delete_spam_account():
...
+    from haystack import connections
+
+    src_backend = connections[ES_INDEX_HELPER].get_backend()
+    dest_backend = connections[ES_INDEX_DEFAULT].get_backend()
+
+    helpers.reindex(
```
akatsoulas: ES docs suggest to use the API instead of the `reindex` helper and this is mostly used for backwards compatibility.

mozillians/settings/base_env_vars.py
```diff
@@ -587,43 +587,26 @@ def _username_algo(email):
...
+    haystack_connections = {
+        'default': {
+            'ENGINE': 'haystack.backends.elasticsearch_backend.ElasticsearchSearchEngine',
+            'URL': es_url,
+            'INDEX_NAME': es_index_name
+        },
+        'helper': {
+            'ENGINE': 'haystack.backends.elasticsearch_backend.ElasticsearchSearchEngine',
+            'URL': es_url,
+            'INDEX_NAME': 'helper_{}'.format(es_index_name)
```
akatsoulas: Probably this is not enough. I am under the impression that haystack will merge the two indexes into one. You need to exclude one of them when you are issuing a reindex command and then copy the contents.

mozillians/users/tasks.py
akatsoulas: FYI the reindex command does not transfer the mappings. We need to make sure that our 2 indexes will have the same mappings.

mozillians/settings/base_env_vars.py
johngian: I am excluding the default one using the ``using`` arg

johngian:
@akatsoulas r?

I refactored how we deal with rebuilding the index because in the previous attempt we only rebuilt data but no mapping.

What I am trying to do now is the following:

* Rebuild a helper index using haystack's commands
* Copy data/mapping/settings to a new index
* Create an alias for that
* Delete old aliases
* Cleanup helper index

This way we have zero downtime (because alias handling is atomic)

mozillians/users/tasks.py
```diff
@@ -319,5 +320,75 @@ def delete_spam_account():
...
+            'index': ES_INDEX_HELPER,
+        },
+        'dest': {
+            'index': ES_INDEX_ALIAS
+        }
+    }
+
+    es_conn.reindex(reindex_query)
+
+    # Link `default` index to alias
```
akatsoulas: Before linking to default, don't we need to reindex the default?

My understanding is that we should perform the following actions:
* Create the helper index
* Create an alias and point it to the helper index.
* Reindex the default index
* Point the alias to the default index
* Remove the helper index

mozillians/users/tasks.py
```diff
@@ -319,5 +320,75 @@ def delete_spam_account():
...
+    alias_index_configuration = {
+        'settings': helper_index[ES_INDEX_HELPER]['settings'],
+        'mappings': helper_index[ES_INDEX_HELPER]['mappings']
+    }
+    es_conn.indices.create(
+        ES_INDEX_ALIAS,
+        alias_index_configuration
+    )
+
+    # Copy documents from `helper` to alias
```
akatsoulas: Do we need to copy the data? Isn't the creation of an alias enough?

johngian: No its not we need to copy the data as well. The alias is where the actual index leaves. Then we just link it to "mozillians" index.

mozillians/users/tasks.py
johngian: No because that way everything gets linked to helper which is disposable.

The process is the following:

* Create the helper index using haystacks command
* Create a replica of this which is the actual index that's going to be used
* Point "mozillians" to be using this alias
* Delete helper
* Delete old aliases

akatsoulas:
Please update the naming schema to better show what's the flow. ALIAS is maybe confusing. CURRENT and TEMP are probably a bit better.

PR #1799, opened 2017-07-26
CIS integration

johngian:


mozillians/common/authbackend.py
```diff
@@ -51,6 +51,13 @@ def filter_users_by_claims(self, claims):
         alternate_email_qs = Q(userprofile__externalaccount=alternate_emails)
         user_q = self.UserModel.objects.filter(primary_email_qs | alternate_email_qs).distinct()
 
+        # Store auth0 user_id in UserProfile
+        if user_q:
+            if user_q.count() == 1:
```
akatsoulas: Why is this needed?

mozillians/users/tasks.py
```diff
@@ -393,3 +394,96 @@ def index_all_profiles():
...
+@task
+def send_userprofile_to_cis(instance_id, **kwargs):
+    from cis.streams import invoke_cis_lambda
+    from nameparser import HumanName
+
+    from mozillians.groups.models import GroupMembership
+    from mozillians.users.models import UserProfile, ExternalAccount
+
+    profile = UserProfile.objects.get(pk=instance_id)
+    hn = HumanName(profile.full_name)
```
akatsoulas: Please replace the `hn` abbreviation with the full name (human_name). It's easier to understand and follow the flow

mozillians/users/tasks.py
```diff
@@ -393,3 +394,96 @@ def index_all_profiles():
         'interactive': False
     }
     call_command('clear_index', **options)
+
+
+@task
+def send_userprofile_to_cis(instance_id, **kwargs):
```
akatsoulas: How/When this task will be triggered?

mozillians/users/tasks.py
akatsoulas: You can save the first/last name to the User object. Right now these fields are empty so this could be a good use for them.

mozillians/users/tasks.py
```diff
@@ -393,3 +394,96 @@ def index_all_profiles():
...
+                'primary': False,
+                'name': 'mozillians-alternate-{}'.format(email.pk)
+            }
+            emails.append(entry)
+
+        return emails
+
+    def _get_uris(profile):
+        accounts = []
+        for account in profile.externalaccount_set.exclude(type=ExternalAccount.TYPE_EMAIL):
```
akatsoulas: Do we want to push all the external accounts that we have down to the CIS or a subset of them? Is every entry in the external accounts verifiable? If not, I would suggest to remove at least those.

mozillians/users/tasks.py
```diff
@@ -393,3 +394,96 @@ def index_all_profiles():
...
+            }
+        ]
+
+        # Alternate emails
+        for email in profile.alternate_emails:
+            entry = {
+                'value': email.identifier,
+                'verified': True,
+                'primary': False,
+                'name': 'mozillians-alternate-{}'.format(email.pk)
```
akatsoulas: Maybe we can use a combination of username and the domain of the email to name each entry. Although the primary key will do the job, probably it's not descriptive enough.

mozillians/users/tasks.py
```diff
@@ -393,3 +394,96 @@ def index_all_profiles():
...
+    from cis.streams import invoke_cis_lambda
+    from nameparser import HumanName
+
+    from mozillians.groups.models import GroupMembership
+    from mozillians.users.models import UserProfile, ExternalAccount
+
+    profile = UserProfile.objects.get(pk=instance_id)
+    hn = HumanName(profile.full_name)
+
+    def _get_emails(profile):
```
akatsoulas: Maybe we can move these helper functions either in a new app or in the common one. Probably makes sense to have all the utilities that mozillians.org will use for CIS in one place.

akatsoulas: Another idea is to have each helper function as a property in each model:

```python
@property
def get_profile_emails(self):
    do_stuff
```

mozillians/common/authbackend.py
johngian: We want to ensure only 1 user instance gets returned in mozilla-django-oidc:

https://github.com/mozilla/mozilla-django-oidc/blob/master/mozilla_django_oidc/auth.py#L152

mozillians/users/tasks.py
johngian: TBD

mozillians/users/tasks.py
johngian: I suggest we send everything for now since at the moment we have no external accounts verified. We can reiterate to add some logic behind verified external accounts in the future.

mozillians/users/tasks.py
johngian: A user might have the same email provider with multiple emails so your suggestion is not really working in all cases. Also the domain information is redundant in the json and we want to keep it as clean as possible.

akatsoulas: the combination of username and domain shall work for every case since each username is unique for the same domain. I am also open to other suggestions. The only reason that I am proposing this it's because it's easier to read for example `mozillians-alternate-nemo-mozilla` and `mozillians-alternate-johngian-mozilla` instead of 'mozillians-alternate-100`. That said, I don't have strong feelings about this

mozillians/users/tasks.py
akatsoulas: I would go the other way around. Let's add the minimum. Eg github. This way we will definitely know if we need more and we can add the logic. If we add them all now, there is a good chance that we might forget to remove something.

mozillians/common/authbackend.py
```diff
@@ -51,6 +51,12 @@ def filter_users_by_claims(self, claims):
         alternate_email_qs = Q(userprofile__externalaccount=alternate_emails)
         user_q = self.UserModel.objects.filter(primary_email_qs | alternate_email_qs).distinct()
 
+        # Store auth0 user_id in UserProfile
+        if user_q.exists() and user_q.count() == 1:
+            profile_id = user_q[0].userprofile.id
+            profile_qs = UserProfile.objects.filter(pk=profile_id)
+            profile_qs.update(auth0_user_id=claims.get('user_id'))
```
akatsoulas: Why don't you just use (a couple of lines below):
```python
request_user.profile.auth0_user_id=claims.get('user_id')
request_user.profile.save() 
```

This way you won't have to perform any of the checks that you have here.

mozillians/users/tasks.py
```diff
@@ -393,3 +396,62 @@ def index_all_profiles():
...
+def send_userprofile_to_cis(instance_id, **kwargs):
+    import boto3
+
+    from cis.encryption import encrypt_payload
+    from mozillians.users.models import UserProfile
+
+    profile = UserProfile.objects.get(pk=instance_id)
+    human_name = HumanName(profile.full_name)
+
+    data = {
```
akatsoulas: I would suggest to make a helper/utility that will get a profile as argument and return all the data. I have a feeling that we will need this in other places too.

mozillians/users/tasks.py
```diff
@@ -393,3 +396,62 @@ def index_all_profiles():
...
+        'SSHFingerprints': [],
+        'PGPFingerprints': [],
+        'authoritativeGroups': []
+    }
+
+    # Invoke lambda
+    profile = json.dumps(data)
+    encrypted_profile = encrypt_payload(profile)
+
+    base64_payload = dict()
```
akatsoulas: Again maybe another utility for this?

mozillians/users/models.py
```diff
@@ -550,6 +553,55 @@ def get_annotated_groups(self):
             groups.append(group)
         return groups
 
+    def get_cis_emails(self):
```
akatsoulas: Ideally the three new methods need tests but since this is going to change a lot and it's just a POC, I will be happy for now with a comment as a placeholder to add the actual tests.

mozillians/users/tasks.py
akatsoulas: both comments about the utilities are nits. We can do this as we proceed.

mozillians/common/authbackend.py
akatsoulas: you missed this :)

mozillians/users/tasks.py
johngian: This is a workaround because there is an issue in the core cis. These functionality is going to be replaced by a single call.

PR #1805, opened 2017-08-02
Automatically remove non-vouched reported profiles.

akatsoulas:


mozillians/users/tasks.py
```diff
@@ -267,55 +265,18 @@ def check_spam_account(instance_id, **kwargs):
...
-        subject = 'Spam profile report {}'.format(datetime.now().strftime('%x'))
-        message = render_to_string('phonebook/emails/admin_spam.txt', {
-            'reports': reports
-        })
-        mail_admins(subject, message)
+    # Delete all unvouched profiles reported as spam by a user
+    reports = AbuseReport.objects.filter(profile__is_vouched=False,
+                                         is_akismet=False,
+                                         reporter__isnull=False)
+    reports.delete()
```
johngian: This way you are deleting the reports, not the profiles.

PR #1809, opened 2017-08-06
Update Discourse URL

flamingspaz:


johngian:
This is going to need a data migration too.

johngian:
@flamingspaz please hold on merging

johngian:
So apparently the datamigration is not required since the identifier per user is going to be the same.

PR #1810, opened 2017-08-09
Improve celery monitoring.

johngian:
* Notify healtchecks.io periodically to make sure that celery/celerybeat is up
  and running


johngian:
@flamingspaz or @comzeradd r?

PR #1814, opened 2017-08-14
Adapt CIS task to server changes.

johngian:
* Add required env vars in settings
* Use ChangeDelegate object instead of deprecated method
* Use sts assume role instead of using hardcoded keys
* Add required pip dependencies


johngian:
@akatsoulas r?

PR #1815, opened 2017-08-14
Add new CDN domain to CSP

flamingspaz:


flamingspaz:
I added the staging CDN domain to this too.

PR #1828, opened 2017-09-11
Only set SES access credentials when we don't want boto to figure it out

flamingspaz:


johngian:
Travis is failing

flamingspaz:
fixed

PR #1829, opened 2017-09-15
Added Welcome Bot config file

hmitsch:


.github/config.yml
```diff
@@ -0,0 +1,21 @@
...
+
+# Comment to be posted to on PRs from first time contributors in your repository
+newPRWelcomeComment: >
+  Thanks for opening this pull request! Please check out our contributing guidelines.
+
+# Configuration for first-pr-merge - https://github.com/behaviorbot/first-pr-merge
+
+# Comment to be posted to on pull requests merged by a first time user
+firstPRMergeComment: >
+  Congrats on merging your first pull request! We here at behaviorbot are proud of you!
```
MichaelKohler: shouldn't that mention Mozilla instead of behaviourbot? Or who exactly is proud? ;)

.github/config.yml
```diff
@@ -0,0 +1,21 @@
...
+newPRWelcomeComment: >
+  Thanks for opening this pull request! Please check out our contributing guidelines.
+
+# Configuration for first-pr-merge - https://github.com/behaviorbot/first-pr-merge
+
+# Comment to be posted to on pull requests merged by a first time user
+firstPRMergeComment: >
+  Congrats on merging your first pull request! We here at behaviorbot are proud of you! 
+
+# It is recommend to include as many gifs and emojis as possible!
```
MichaelKohler: Why don't we? ;)

.github/config.yml
```diff
@@ -0,0 +1,21 @@
+# Configuration for welcome - https://github.com/behaviorbot/welcome
+
+# Configuration for new-issue-welcome - https://github.com/behaviorbot/new-issue-welcome
+
+# Comment to be posted to on first time issues
+newIssueWelcomeComment: >
+  Thanks for opening your first issue here! Be sure to follow the issue template!
```
MichaelKohler: Isn't it somehow confusing to mention the issue template if somebody just filed an issue which had the issue template in it?

MichaelKohler: On another note, as mozillians does not have issues, this might be obsolete anyway.

hmitsch:
@MichaelKohler I think I addressed most of your feedback in the second commit.

.github/config.yml
hmitsch: Changed the text to something that is generically meaningful to ParSys, even though it will not be needed for the mozillians repo.

.github/config.yml
hmitsch: Totally. Fixed.

.github/config.yml
hmitsch: I don't know how to add emoji. Do you know how that works? Would I only have to write `:heart:` (for example)? 
Do you have any GIFs you'd like to see included?

Mte90:
Also probably we don't need the text for issue because this repo doesn't have them enabled and in case that bot now have a default text (I opened a ticket for that and already fixed by the author)

hmitsch:
@Mte90 I would like to keep the `newIssueWelcomeComment` text because I hope to reuse it with other ParSys repos. :-)

Mte90:
Good point!

.github/config.yml
darkvn79: meaningful to ParSys

.github/config.yml
darkvn79: omg i don't know , emoji auto add to my phone

.github/config.yml
darkvn79: thank you , i not sure my text

johngian:
Closing this one after coordinating with @hmitsch

PR #1831, opened 2017-10-03
Integrate mozillians with CIS and multiple auth0 user ids

johngian:


johngian:
@akatsoulas ^

mozillians/common/authbackend.py
```diff
@@ -50,24 +50,41 @@ def filter_users_by_claims(self, claims):
...
+            idp_q.exclude(auth0_user_id=auth0_user_id).update(primary=False)
 
-        # In this case we have a registered user who is adding a secondary email
+            # Mark current `user_id` as `primary=True`
+            idp_q.filter(auth0_user_id=auth0_user_id).update(primary=True)
+
+        # Add alternate email
         if request_user.is_authenticated():
-            if not user_q:
+            email_q = self.UserModel.objects.filter(primary_email_qs | alternate_email_qs)
```
akatsoulas: Should we still allow users to add new emails in general or allow only Github/Google?

mozillians/users/migrations/0021_auto_20171003_0319.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
akatsoulas: Isn't this identical to the file above?

mozillians/users/models.py
```diff
@@ -596,13 +594,35 @@ def get_cis_uris(self):
...
         memberships = GroupMembership.objects.filter(
             userprofile=self,
-            status=GroupMembership.MEMBER
+            status=GroupMembership.MEMBER,
+            group__name__in=CIS_GROUPS
         )
         groups = ['mozilliansorg_{}'.format(m.group.name) for m in memberships]
         return groups
 
+    def get_cis_tags(self):
```
akatsoulas: Let's add a few tests for both this method and the `get_cis_groups`.

mozillians/users/models.py
```diff
@@ -627,6 +647,49 @@ def save(self, *args, **kwargs):
...
+                            default=PROVIDER_PASSWORDLESS,
+                            max_length=50,
+                            blank=False)
+    # Auth0 required data
+    auth0_user_id = models.CharField(max_length=1024, default='', blank=True)
+    primary = models.BooleanField(default=False)
+    created = models.DateTimeField(auto_now_add=True)
+    updated = models.DateTimeField(auto_now=True)
+
+    def get_provider_type(self):
```
akatsoulas: Docstrings pls!

mozillians/users/models.py
```diff
@@ -627,6 +647,49 @@ def save(self, *args, **kwargs):
...
+        if 'github|' in self.auth0_user_id:
+            return self.PROVIDER_GITHUB
+
+        if 'google-oauth2|' in self.auth0_user_id:
+            return self.PROVIDER_GOOGLE
+
+        if 'email|' in self.auth0_user_id:
+            return self.PROVIDER_PASSWORDLESS
+
+    def save(self, *args, **kwargs):
```
akatsoulas: same here.

mozillians/users/tasks.py
```diff
@@ -413,8 +385,40 @@ def send_userprofile_to_cis(instance_id, **kwargs):
...
+            'SSHFingerprints': [],
+            'PGPFingerprints': [],
+            'authoritativeGroups': []
+        }
 
-    return result
+        cis_change = ChangeDelegate(publisher, {}, data)
+        cis_change.boto_session = session
+        result = cis_change.send()
+        results.append(result)
```
akatsoulas: How is the upgrade process handled (eg from passwordless -> github -> ldap)?

mozillians/users/models.py
johngian: Done

mozillians/users/models.py
johngian: Done

mozillians/common/authbackend.py
johngian: I suggest allow adding in general.

johngian:
@akatsoulas PR updated with:

* Update strategy handling
* Fix IdpProfiles on user creation

Also tested it with CIS and it looks like it triggers lambda as expected.

mozillians/common/authbackend.py
```diff
@@ -50,24 +63,68 @@ def filter_users_by_claims(self, claims):
...
         alternate_email_qs = Q(userprofile__externalaccount=alternate_emails)
-        user_q = self.UserModel.objects.filter(primary_email_qs | alternate_email_qs).distinct()
 
-        # Store auth0 user_id in UserProfile
-        if user_q.exists() and user_q.count() == 1:
-            profile_id = user_q[0].userprofile.id
-            profile_qs = UserProfile.objects.filter(pk=profile_id)
-            profile_qs.update(auth0_user_id=claims.get('user_id'))
+        # Allow logins only using primary email
+        user_q = self.UserModel.objects.filter(primary_email_qs).distinct()
```
akatsoulas: You don't need nor `distinct` or `Q` object here. You can filter directly on the `email` field of the `User` object.

mozillians/common/authbackend.py
```diff
@@ -50,24 +63,68 @@ def filter_users_by_claims(self, claims):
...
 
-        # In this case we have a registered user who is adding a secondary email
+        if user_q.exists() and user_q.count() == 1:
```
akatsoulas: Can two users have the same email?

mozillians/common/authbackend.py
```diff
@@ -50,24 +63,68 @@ def filter_users_by_claims(self, claims):
...
+            # Get or create an IdpProfile for this user
+            profile = user_q[0].userprofile
+            auth0_user_id = claims.get('user_id')
+
+            # Get current_idp
+            current_idp = get_object_or_none(IdpProfile, profile=profile, primary=True)
```
akatsoulas: Since you are creating an IdpProfile in the `create_user`, this should never return `None`.

mozillians/common/authbackend.py
```diff
@@ -50,24 +63,68 @@ def filter_users_by_claims(self, claims):
...
+
+            # Get or create new `user_id`
+            obj, created = IdpProfile.objects.get_or_create(
+                profile=profile,
+                auth0_user_id=auth0_user_id)
+
+            if current_idp:
+                allowed_idp_flows = {
+                    IdpProfile.PROVIDER_PASSWORDLESS: [
```
akatsoulas: Shall we move this on the top of the file or in `__init__` in ALLOWED_IDP_FLOWS?

mozillians/users/tasks.py
```diff
@@ -413,8 +385,40 @@ def send_userprofile_to_cis(instance_id, **kwargs):
         'id': settings.CIS_PUBLISHER_NAME
     }
 
-    cis_change = ChangeDelegate(publisher, {}, data)
-    cis_change.boto_session = session
-    result = cis_change.send()
+    results = []
+    for idp in profile.idp_profiles.all():
```
akatsoulas: This change will push data for all the Idp profiles of a user. Do we want to do this or just push data for the primary?

mozillians/users/tasks.py
akatsoulas: When there is an upgrade, shouldn't we remove the access groups from the old profile and push them to the new one? Is this going to be handled in another PR?

mozillians/users/tasks.py
johngian: for all. Then ``get_cis_groups`` will now when to wipe groups and when not to

mozillians/common/authbackend.py
johngian: User might have never logged in after the changes. It's better to handle this gracefully.

mozillians/common/authbackend.py
```diff
@@ -50,24 +78,54 @@ def filter_users_by_claims(self, claims):
...
+
+            # Get or create new `user_id`
+            obj, created = IdpProfile.objects.get_or_create(
+                profile=profile,
+                auth0_user_id=auth0_user_id)
 
-        # In this case we have a registered user who is adding a secondary email
+            if current_idp:
+
+                if obj.type not in ALLOWED_IDP_FLOWS[current_idp.type]:
```
akatsoulas: How is this ensuring that a user cannot go back? eg from ldap to passwordless

johngian: Because for example ``passwordless`` is not going to be a value in the list with key ``ldap``.

mozillians/common/authbackend.py
```diff
@@ -50,24 +78,54 @@ def filter_users_by_claims(self, claims):
...
+        if user_q.exists():
+            # Get or create an IdpProfile for this user
+            profile = user_q[0].userprofile
+            auth0_user_id = claims.get('user_id')
+
+            # Get current_idp
+            current_idp = get_object_or_none(IdpProfile, profile=profile, primary=True)
+
+            # Get or create new `user_id`
+            obj, created = IdpProfile.objects.get_or_create(
```
akatsoulas: nit: s/created/_ since you are not using it anywhere

mozillians/common/authbackend.py
akatsoulas: Pls add comments to describe the flow. Also I would suggest to add a couple of tests showcasing that a user cannot have a "backwards" flow

mozillians/users/models.py
```diff
@@ -594,15 +592,44 @@ def get_cis_uris(self):
...
+        ordering = [
+            IdpProfile.PROVIDER_PASSWORDLESS,
+            IdpProfile.PROVIDER_GOOGLE,
+            IdpProfile.PROVIDER_GITHUB,
+            IdpProfile.PROVIDER_LDAP
+        ]
+
+        idps.sort(key=ordering.index)
+
+        if idp.type != idps[-1]:
```
akatsoulas: comments pls :)

PR #1834, opened 2017-10-09
Add an env var for axes reverse proxy.

akatsoulas:
* Remove duplicate configuration setting.


mozillians/settings.py
```diff
@@ -506,9 +506,6 @@ def _lazy_haystack_setup():
     }
 ]
 
-# Proxy configuration
-SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
```
johngian: Why are you removing this entry?

akatsoulas: We have a couple of those... :P

PR #1839, opened 2017-10-10
Account verification

akatsoulas:


mozillians/phonebook/views.py
```diff
@@ -535,3 +543,119 @@ def get_context_data(self, **kwargs):
...
+        if nonce:
+            # Make sure that nonce is not used twice
+            del request.session['oidc_verify_nonce']
+
+        if 'code' not in request.GET or 'state' not in request.GET:
+            msg = 'Account verification failed: `code or state is missing.`'
+            messages.error(request, msg)
+
+        if 'oidc_verify_state' not in request.session:
+            msg = 'Account verification failed: `oidc_verify_state not in session.`'
```
johngian: Let's avoid adding technical terms in the user messages.
Just ``Something went wrong`` is better. Also it reduces the complexity of these if cases.

FYI users don't understand what state, code etc mean

mozillians/phonebook/views.py
```diff
@@ -535,3 +543,119 @@ def get_context_data(self, **kwargs):
...
+
+        msg = ''
+        # This is a difference nonce than the one used to login!
+        nonce = request.session.get('oidc_verify_nonce')
+        if nonce:
+            # Make sure that nonce is not used twice
+            del request.session['oidc_verify_nonce']
+
+        if 'code' not in request.GET or 'state' not in request.GET:
+            msg = 'Account verification failed: `code or state is missing.`'
```
johngian: Same here

mozillians/phonebook/views.py
```diff
@@ -535,3 +543,119 @@ def get_context_data(self, **kwargs):
...
+            messages.error(request, msg)
+
+        if 'oidc_verify_state' not in request.session:
+            msg = 'Account verification failed: `oidc_verify_state not in session.`'
+            messages.error(request, msg)
+
+        if request.GET['state'] != request.session['oidc_verify_state']:
+            msg = 'Account verification failed: `oidc_verify_state does not match callback state.`'
```
johngian: Same here

mozillians/settings.py
```diff
@@ -620,6 +624,7 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
                          '^/api/',
                          '^/oidc/authenticate/',
                          '^/oidc/callback/',
+                         '^/verify/identity/callback/',
```
johngian: User needs to be authenticated to use this callback. Let's avoid adding this here.

mozillians/phonebook/views.py
```diff
@@ -535,3 +543,133 @@ def get_context_data(self, **kwargs):
...
+
+        profile = request.user.userprofile
+        # This is a difference nonce than the one used to login!
+        nonce = request.session.get('oidc_verify_nonce')
+        if nonce:
+            # Make sure that nonce is not used twice
+            del request.session['oidc_verify_nonce']
+
+        # Check for all possible errors and display a message to the user.
+        if ('code' not in request.GET or 'state' not in request.GET or
```
johngian: nit: I would change this to 
```
errors = [
    'code' not in request.GET,
    'state' not in ...,
    ...
]

if any(errors):
    ...
```

mozillians/phonebook/views.py
```diff
@@ -535,3 +543,133 @@ def get_context_data(self, **kwargs):
...
+                msg = 'Account verification failed: `Email is not verified`.'
+                messages.error(request, msg)
+                return redirect('phonebook:profile_edit')
+
+            # Save the new identity to the IdpProfile
+            user_q = {
+                'profile': profile,
+                'auth0_user_id': user_info['sub'],
+                'email': user_info['email']
+            }
```
johngian: what about the ``primary`` attribute ?

mozillians/users/models.py
```diff
@@ -678,8 +678,10 @@ class IdpProfile(models.Model):
     # Auth0 required data
     auth0_user_id = models.CharField(max_length=1024, default='', blank=True)
     primary = models.BooleanField(default=False)
+    email = models.EmailField(blank=True, default='')
     created = models.DateTimeField(auto_now_add=True)
     updated = models.DateTimeField(auto_now=True)
+    privacy = models.PositiveIntegerField(default=MOZILLIANS, choices=PRIVACY_CHOICES)
```
johngian: This needs some change in the template too right?

mozillians/phonebook/views.py
akatsoulas: Adding a new identity doesn't necessarily mean that is a primary one. Eg you can have ldap as primary but you want to add a Google one

johngian: No i mean we are using the default value right? It's false.

mozillians/users/models.py
akatsoulas: Right now, no it doesn't because we don't expose this model anywhere in the front-end. That said, I am already in the process of swapping the alternate emails formset with a new one. I will tackle that in a following PR.

mozillians/phonebook/views.py
akatsoulas: Oh, yes that's correct

PR #1850, opened 2017-10-13
Check that an identity is tied to a single profile.

akatsoulas:


akatsoulas:
@johngian r?

mozillians/phonebook/views.py
```diff
@@ -661,19 +661,25 @@ def get(self, request):
...
 
-            # Save the new identity to the IdpProfile
             user_q = {
-                'profile': profile,
                 'auth0_user_id': user_info['sub'],
                 'email': user_info['email']
             }
 
+            # Check that the identity doesn't exist in another profile
+            if IdpProfile.objects.filter(**user_q).exists():
```
johngian: This needs to be an OR not an AND. We want:

* The email to not exist to another IdpProfile

or

* The user_id to not exist to another IdpProfile

johngian:
r+wc

mozillians/phonebook/views.py
akatsoulas: this email is the one returned by Auth0 and not the one in the User object. It's not possible to have a combination of the same auth0_id and email in multiple identities. 

On the other hand is totally valid to have the same email in multiple Auth0 IDs. Eg a user that has the Google email used in GitHub.

akatsoulas: That said, if the above query returns True, then it means that a user already had added those. We don't care which profile in mozillians holds them. We just want those to be unique in the DB.

PR #1853, opened 2017-10-16
Check if a profile exists in the CIS task.

akatsoulas:


johngian:
You are not sending an empty list to CIS. This changes the celery results + error handling if user does not exist.

Please update the commit message.

r+wc

PR #1906, opened 2017-10-19
Allow users to login with a verified MFAed account.

akatsoulas:


mozillians/common/authbackend.py
```diff
@@ -76,8 +76,19 @@ def create_user(self, claims):
...
+        user_q = super(MozilliansAuthBackend, self).filter_users_by_claims(claims)
         self.claims = claims
-        return super(MozilliansAuthBackend, self).filter_users_by_claims(claims)
+
+        # The email search did not return a match, let's look in the Identity Profiles
+        if not user_q:
+            idp_q = IdpProfile.objects.filter(email=claims.get('email'),
+                                              auth0_user_id=claims.get('user_id'))
+
+            if len(idp_q) == 1:
```
johngian: Instead of evaluation to list and then apply ``len`` let's just use ``idp_q.count()``

mozillians/common/authbackend.py
```diff
@@ -76,8 +76,19 @@ def create_user(self, claims):
...
     def filter_users_by_claims(self, claims):
         """Override default method to store claims."""
+
+        user_q = super(MozilliansAuthBackend, self).filter_users_by_claims(claims)
         self.claims = claims
-        return super(MozilliansAuthBackend, self).filter_users_by_claims(claims)
+
+        # The email search did not return a match, let's look in the Identity Profiles
+        if not user_q:
+            idp_q = IdpProfile.objects.filter(email=claims.get('email'),
```
johngian: This way we are allowing logging in with multiple emails.

mozillians/users/models.py
```diff
@@ -697,7 +696,7 @@ def get_provider_type(self):
         if 'email|' in self.auth0_user_id:
             return self.PROVIDER_PASSWORDLESS
 
-        return ''
+        return self.PROVIDER_PASSWORDLESS
```
johngian: This is not correct. In case of malformed data or in case something goes wrong we are going to have wrong data in the DB.

johngian: You can use ``None`` instead.

johngian:
Since we are going to change the way we order IDP hierarchy please fix this here too:
https://github.com/mozilla/mozillians/blob/master/mozillians/users/models.py#L600

mozillians/common/authbackend.py
```diff
@@ -104,9 +104,12 @@ def check_authentication_method(self, user):
 
         if current_idp:
 
-            if obj.type not in ALLOWED_IDP_FLOWS[current_idp.type]:
+            # Do not allow downgrades.
+            if obj.type > current_idp.type:
                 msg = u'Please use one of the following authentication methods: {}'
-                methods = ', '.join(ALLOWED_IDP_FLOWS[current_idp.type])
+                # convert the tuple to a dict to easily get the values
+                provider_types = dict(IdpProfile.PROVIDER_TYPES)
```
johngian: Instead of converting you can iterate over the tuple and select the element using indices.

mozillians/phonebook/views.py
```diff
@@ -674,7 +674,17 @@ def get(self, request):
 
             # Save the new identity to the IdpProfile
             user_q['profile'] = profile
-            _, created = IdpProfile.objects.get_or_create(**user_q)
+            idp, created = IdpProfile.objects.get_or_create(**user_q)
+
+            current_idp = get_object_or_none(IdpProfile, profile=profile, primary=True)
+            # The new identity is stronger than the one currently used. Let's swap
+            if current_idp and current_idp.type > idp.type:
```
johngian: Here you check if the current idp is stronger than the new one. Shouldn't it be the other way around?

akatsoulas: the smaller the number the stronger the authentication method so this should be correct. Can you verify pls?

mozillians/users/models.py
```diff
@@ -671,10 +663,10 @@ class IdpProfile(models.Model):
 
     )
     profile = models.ForeignKey(UserProfile, related_name='idp_profiles')
-    type = models.CharField(choices=PROVIDER_TYPES,
-                            default=PROVIDER_PASSWORDLESS,
-                            max_length=50,
-                            blank=False)
+    type = models.IntegerField(choices=PROVIDER_TYPES,
+                               default=PROVIDER_PASSWORDLESS,
```
johngian: This should default to None

johngian:
r+wc

PR #1907, opened 2017-10-20
Check if IDP is MFA-ed when populating cis groups.

johngian:


mozillians/users/models.py
```diff
@@ -691,6 +691,11 @@ def get_provider_type(self):
 
         return None
 
+    def is_mfa(self):
+        """Helper method to check if IdpProfile is MFA-ed"""
+        mfa_providers = [self.PROVIDER_GITHUB, self.PROVIDER_LDAP]
+        return self.type in mfa_providers
```
akatsoulas: no need to assign a variable for this.

PR #1909, opened 2017-10-20
Check if an email belongs to another user before making it primary.

akatsoulas:


mozillians/phonebook/tests/test_identities.py
```diff
@@ -0,0 +1,168 @@
...
+        }
+        with self.login(user) as client:
+            session = client.session
+            session['oidc_verify_nonce'] = 'nonce'
+            session['oidc_verify_state'] = 'state'
+            session.save()
+            client.get(self.url, self.get_data, follow=True)
+            new_idp_profile = IdpProfile.objects.get(email='bar@example.com')
+            eq_(new_idp_profile.primary, False)
+            msg = 'Account successfully verified.'
```
akatsoulas: Note to self: also assert redirects.

PR #1911, opened 2017-10-23
Replace Alternate emails with Idp Profile in settings.

akatsoulas:
* Add new urls.
* Add new privacy field for emails and Idp Profiles.
* Update JS to handle the new field.
* Update phonebook forms.


akatsoulas:
Tests will fail but I opened the PR to start a review on the logic applied so far.

mozillians/jinja2/phonebook/includes/profile_edit_basic.html
```diff
@@ -99,10 +99,10 @@
...
     <div class="panel panel-default">
-      <div class="panel-heading">{{ _('Emails') }}</div>
+      <div class="panel-heading">{{ _('Profile Identities') }}</div>
       <div class="panel-body">
         <fieldset>
-          <label class="email-label">{{ _('Primary e-mail address') }}</label>
+          <label class="identity-profile-label">{{ _('Login Identity') }}</label>
           <div class="row">
             <div class="col-md-8 col-sm-6">
               <span class="email">{{ email_privacy_form.instance.email }}</span>
```
akatsoulas: FYI, this needs changing to the primary IDP contact identity. I believe this is the last change standing (and the tests of course).

mozillians/jinja2/phonebook/includes/profile_edit_basic.html
```diff
@@ -99,36 +99,43 @@
...
-          <label class="email-label">{{ _('Primary e-mail address') }}</label>
+          <label class="identity-profile-label">{{ _('Login Identity') }}</label>
           <div class="row">
             <div class="col-md-8 col-sm-6">
-              <span class="email">{{ email_privacy_form.instance.email }}</span>
-            </div>
-            <div class="col-md-4 col-sm-6 text-right">
-              {{ privacy_field(email_privacy_form.privacy_email) }}
+              <span class="email">
+                {{ idp_profile_form.instance.get_type_display() }}
```
johngian: I am not sure how this is rendered but do we really want to have this span
```
<span>Github foobar</span>
```

mozillians/jinja2/phonebook/profile.html
```diff
@@ -302,16 +303,17 @@ <h1 class="p-name">
...
-              {% for email in profile.alternate_emails %}
+              {% for idp in profile.idp_profiles %}
                 <li>
+                  {{ idp.get_type_display() }} -
                   {% if user != shown_user %}
-                    {{ email.identifier|urlize }}
+                    {{ idp.email|urlize }}
                   {% else %}
-                    {{ email.identifier }}
+                    {{ idp.email }}
```
johngian: We should consider url-izing this email as well.

mozillians/phonebook/views.py
```diff
@@ -307,48 +305,49 @@ def get_request_data(form):
...
 @never_cache
-def delete_email(request, email_pk):
+def delete_identity(request, identity_pk):
     """Delete alternate email address."""
     user = User.objects.get(pk=request.user.id)
     profile = user.userprofile
 
     # Only email owner can delete emails
-    if not ExternalAccount.objects.filter(user=profile, pk=email_pk).exists():
+    idp_query = IdpProfile.objects.filter(profile=profile, pk=identity_pk)
```
johngian: Why do we need both ``profile=profile`` and ``pk=identity_pk`` ?

mozillians/phonebook/views.py
```diff
@@ -307,48 +305,49 @@ def get_request_data(form):
...
 @allow_unvouched
 @never_cache
-def change_primary_email(request, email_pk):
+def change_primary_contact_identity(request, identity_pk):
     """Change primary email address."""
     user = User.objects.get(pk=request.user.id)
     profile = user.userprofile
-    alternate_emails = ExternalAccount.objects.filter(user=profile,
-                                                      type=ExternalAccount.TYPE_EMAIL)
+    alternate_identities = IdpProfile.objects.filter(profile=profile)
```
johngian: I think that here you should filter ``primary=False``

mozillians/phonebook/views.py
```diff
@@ -307,48 +305,49 @@ def get_request_data(form):
...
 
     # Only email owner can change primary email
-    if not alternate_emails.filter(pk=email_pk).exists():
+    if not alternate_identities.filter(pk=identity_pk).exists():
         raise Http404()
 
-    alternate_email = alternate_emails.get(pk=email_pk)
-    primary_email = user.email
+    if alternate_identities.filter(primary_contact_identity=True).exists():
```
johngian: You should be able to change your primary contact identity regardless if you already have one.

mozillians/phonebook/views.py
```diff
@@ -307,48 +305,49 @@ def get_request_data(form):
...
-    # Turn primary email to alternate
-    alternate_email.identifier = primary_email
-
-    with transaction.atomic():
-        user.save()
-        alternate_email.save()
-    # Notify Basket about this change
-    update_email_in_basket.delay(primary_email, user.email)
-    send_userprofile_to_cis.delay(profile.pk)
+        msg = _(u'Primary Contact Identity successfully updated.')
```
johngian: Here you are dropping the basket code that we still need

mozillians/users/models.py
```diff
@@ -273,6 +277,12 @@ def _alternate_emails(self):
         return self._filter_accounts_privacy(accounts)
 
     @property
+    def _identity_profiles(self):
+        _getattr = (lambda x: super(UserProfile, self).__getattribute__(x))
+        accounts = _getattr('idp_profiles').filter(primary=False)
```
johngian: Why are you filtering only ``primary=False``?

mozillians/phonebook/views.py
akatsoulas: Because you could pass an id to the view that doesn't belong to your profile.

mozillians/phonebook/views.py
akatsoulas: If you do that, you exclude the primary from the queryset. We need that to update it to `False`.

mozillians/jinja2/phonebook/includes/profile_edit_basic.html
akatsoulas: I would go with `yes` for now. If we don't add that we can end up having a list of the same emails. It's not ideal but I would suggest to tackle this on the second iteration (see the upcoming UI update)

mozillians/phonebook/views.py
akatsoulas: As long as there is one `IdpProfile` you will always have a default identity. This is for the users that have zero identities.

johngian:
```
> grep -R "PRIVILEGED

mozillians/phonebook/tests/test_views/test_views_misc.py:from mozillians.users.managers import MOZILLIANS, PRIVILEGED, PUBLIC
mozillians/phonebook/tests/test_views/test_views_misc.py:        data['privacy_tshirt'] = PRIVILEGED

```

This shouldn't exist since we are removing all reference to ``PRIVILEGED``.

mozillians/phonebook/views.py
johngian: duh!

PR #1914, opened 2017-10-25
Use primary contact

johngian:


mozillians/users/models.py
```diff
@@ -289,7 +289,17 @@ def _languages(self):
...
         _getattr = (lambda x: super(UserProfile, self).__getattribute__(x))
+
         privacy_fields = UserProfile.privacy_fields()
+
+        # Try IDP primary contact first
+        if self.idp_profiles.exists():
+            if self._privacy_level and _getattr('privacy_idp_profile') < self._privacy_level:
+                idp_profiles = privacy_fields['identity_profiles']
+                return idp_profiles.filter(primary_contact_identity=True)
+            return _getattr('user').email
```
akatsoulas: Can you add a couple of unittests to showcase what you are doing here?

mozillians/users/models.py
```diff
@@ -289,7 +289,17 @@ def _languages(self):
...
     def _primary_email(self):
         _getattr = (lambda x: super(UserProfile, self).__getattribute__(x))
+
         privacy_fields = UserProfile.privacy_fields()
+
+        # Try IDP primary contact first
+        if self.idp_profiles.exists():
+            if self._privacy_level and _getattr('privacy_idp_profile') < self._privacy_level:
+                idp_profiles = privacy_fields['identity_profiles']
+                return idp_profiles.filter(primary_contact_identity=True)
```
akatsoulas: This will return an IdpProfile in a QuerySet and not the actual email

mozillians/users/models.py
akatsoulas: Isn't this line going to return the `user.email`? If yes why to we need the lines below that?

mozillians/users/models.py
```diff
@@ -297,10 +297,23 @@ def _languages(self):
...
+
         privacy_fields = UserProfile.privacy_fields()
-        if self._privacy_level and _getattr('privacy_email') < self._privacy_level:
-            email = privacy_fields['email']
-            return email
+
+        if self._privacy_level:
+            # Try IDP contact first
+            if self.idp_profiles.exists():
+                contact_ids = self.identity_profiles.filter(primary_contact_identity=True)
```
akatsoulas: The `identity_profiles` manager is privacy aware, right?

mozillians/users/models.py
```diff
@@ -297,10 +297,23 @@ def _languages(self):
...
+                if contact_ids.exists():
+                    return contact_ids[0].email
+
+            # Fallback to user.email
+            if _getattr('privacy_email') < self._privacy_level:
```
akatsoulas: This will return the email from the profile and not the `user.email`, right? If this is intended please update the comment.

mozillians/users/models.py
```diff
@@ -297,10 +297,23 @@ def _languages(self):
...
+                return privacy_fields['email']
+
+        # In case we don't have a privacy aware attribute access
+        if self.idp_profiles.exists():
+            return self.idp_profiles.filter(primary_contact_identity=True)[0].email
```
akatsoulas: What if we don't have a primary contact identity in the Idp profiles?

akatsoulas: if we don't have a privacy_aware attribute shouldn't we return an empty string?

mozillians/users/models.py
johngian: yes

mozillians/users/models.py
johngian: No this will return the privacy aware ``user.email``. It's either empty or ``user.email``.

mozillians/users/models.py
johngian: If we don't have a primary contact identity it should return the privacy aware ``user.email``.

mozillians/users/tests/test_models.py
```diff
@@ -611,6 +611,69 @@ def test_urls_contain_identifiers(self):
                 ok_('{identifier}' in account['url'])
 
 
+class EmailAttributeTests(TestCase):
+    def test_existing_idp_privacy_unaware(self):
+        profile = UserFactory.create(email='foo@foo.com').userprofile
```
akatsoulas: Shouldn't this be something else to showcase that the email comes from the Idp?

mozillians/users/tests/test_models.py
```diff
@@ -611,6 +611,69 @@ def test_urls_contain_identifiers(self):
...
+            email='foo@bar.com',
+            primary=True,
+            primary_contact_identity=True,
+            privacy=PUBLIC
+        )
+
+        eq_(profile.email, 'foo@bar.com')
+
+    def test_existing_idp_privacy_allowed(self):
+        profile = UserFactory.create(email='foo@foo.com').userprofile
```
akatsoulas: same here

akatsoulas: same here

PR #1926, opened 2017-10-30
Revert "Update API to expose/filter IdpProfiles instead of ExternalAccounts."

johngian:
Reverts mozilla/mozillians#1917


johngian:
@akatsoulas r?

PR #1927, opened 2017-10-30
Workaround for alternate emails/idp API compatibility.

johngian:


johngian:
@akatsoulas r?

mozillians/users/models.py
```diff
@@ -275,6 +276,28 @@ def _alternate_emails(self):
...
+    def _api_alternate_emails(self):
+        legacy_emails_qs = self._alternate_emails
+        idp_qs = self._identity_profiles
+
+        e_exclude = []
+        idp_exclude = []
+
+        for e in legacy_emails_qs:
+            if idp_qs.filter(email=e.identifier, privacy__gte=e.privacy).exists():
+                e_exclude.append(e.id)
```
akatsoulas: nit: we can make this a bit more pythonic with something like:
```python
e_exclude = [e.id for e in legacy_emails_qs if idp_qs.filter(query_here).exists()]
```

fyi, I have not tested this.

mozillians/users/models.py
```diff
@@ -275,6 +276,28 @@ def _alternate_emails(self):
...
+
+        legacy_emails_qs = legacy_emails_qs.exclude(id__in=e_exclude)
+
+        for i in idp_qs:
+            if legacy_emails_qs.filter(identifier=i.email, privacy__gte=i.privacy).exists():
```
akatsoulas: same here

PR #1928, opened 2017-10-31
Fix API compatibility on email filtering

johngian:


mozillians/users/api/v2.py
```diff
@@ -203,9 +203,18 @@ class Meta:
...
+            type=ExternalAccount.TYPE_EMAIL,
+            identifier=value
+        )
+
+        idp_qs = IdpProfile.objects.filter(
+            email=value
+        )
+
+        legacy_users = legacy_qs.values_list('user__id', flat=True)
+        idp_users = idp_qs.values_list('profile__id', flat=True)
```
akatsoulas: Shouldn't this be `distinct()`?

johngian: We are applying distinct later so i think we are good.

PR #1935, opened 2017-11-01
Add a message when there is an account upgrade.

akatsoulas:


mozillians/phonebook/views.py
```diff
@@ -691,10 +692,13 @@ def get(self, request):
                 # Also update the primary email of the user
                 update_email_in_basket(profile.user.email, idp.email)
                 User.objects.filter(pk=profile.user.id).update(email=idp.email)
+                append_msg = ' You need to use this identity the next time you will login.'
 
             send_userprofile_to_cis.delay(profile.pk)
             if created:
                 msg = 'Account successfully verified.'
+                if append_msg:
```
johngian: ``append_msg`` is empty by default so you can always append instead of checking.

PR #1960, opened 2017-11-15
Add access groups through the admin interface.

akatsoulas:


johngian:
This PR also needs a data migration to mark the current access groups as access groups.

PR #1963, opened 2017-11-15
Do not use update when marking an IdpProfile as primary.

akatsoulas:


johngian:
@akatsoulas is this still valid? should i go ahead and delete it?

johngian:
I don't like the idea of merging something that we don't know why it works. Everything we do here is already atomic and we don't have any sort of async access. If something fails we are not going to send anything. I am leaning towards to not merge this one.

akatsoulas:
Can you elaborate on how this is already atomic?

PR #1973, opened 2017-12-06
Notify CIS when a profile is deleted.

akatsoulas:


mozillians/users/tasks.py
```diff
@@ -367,6 +367,8 @@ def send_userprofile_to_cis(instance_id, **kwargs):
     from cis.publisher import ChangeDelegate
     from mozillians.users.models import IdpProfile, UserProfile
 
+    groups = kwargs.pop('groups', None)
```
johngian: Instead of using ``pop`` let's add it as an optional argument that defaults to ``None``.

mozillians/users/signals.py
```diff
@@ -5,7 +5,8 @@
 from django.contrib.auth.models import User
 
 from mozillians.users.models import ExternalAccount, UserProfile, Vouch
-from mozillians.users.tasks import subscribe_user_to_basket, unsubscribe_from_basket_task
+from mozillians.users.tasks import (send_profile_to_cis, subscribe_user_to_basket,
```
johngian: ``send_userprofile_to_cis``

mozillians/users/tasks.py
```diff
@@ -361,20 +360,16 @@ def index_all_profiles():
 
 
 @task
-def send_userprofile_to_cis(instance_id, **kwargs):
+def send_userprofile_to_cis(instance_id='', profile_results=[], **kwargs):
```
johngian: Instead of ``instance_id=''`` use ``None``.

PR #1977, opened 2017-12-06
Do not run lambda calls in tests.

johngian:


johngian:
@akatsoulas r? This is a regression from #1973

PR #1980, opened 2017-12-07
Update autocomplete query

akatsoulas:


mozillians/groups/forms.py
```diff
@@ -141,6 +141,13 @@ def __init__(self, *args, **kwargs):
         if self.instance.pk:
             self.initial['invites'] = []
 
+            # If the group is a TAG or is the NDA we need to query all the vouched
+            # mozillians and not only staff or NDA members.
+            if not self.instance.is_access_group or self.instance.name == 'nda':
```
johngian: please use ``settings.NDA_GROUP`` instead

mozillians/groups/forms.py
```diff
@@ -141,6 +141,13 @@ def __init__(self, *args, **kwargs):
...
+                self.fields['invites'].widget = autocomplete.ModelSelect2Multiple(
```
johngian: Since we are dynamically adding the widget i suggest we remove the entry from ``Meta`` and add it here for consistency.

mozillians/users/views.py
```diff
@@ -120,7 +120,12 @@ def get_queryset(self):
 
         query = reduce(or_, queries)
 
-        return UserProfile.objects.filter(query).distinct()
+        qs = UserProfile.objects.filter(query).distinct()
+        if self.q:
+            qs = qs.filter(Q(full_name__icontains=self.q) |
+                           Q(user__email__icontains=self.q) |
```
johngian: Let's remove this filter since it doesn't really work all the times after the ``IdpProfile`` email changes.

mozillians/users/views.py
```diff
@@ -135,7 +140,12 @@ def get_queryset(self):
         ).values_list('userprofile__pk', flat=True)
 
         query = Q(pk__in=staff_ids) | Q(pk__in=nda_members_ids)
-        return UserProfile.objects.filter(query).distinct()
+        qs = UserProfile.objects.filter(query).distinct()
+        if self.q:
+            qs = qs.filter(Q(full_name__icontains=self.q) |
+                           Q(user__email__icontains=self.q) |
```
johngian: Same here.

PR #1984, opened 2017-12-08
Only allow employees and NDA members to request to join reviewed acce…

akatsoulas:
…ss group.

### DIFF
--- a/mozillians/groups/models.py
+++ b/mozillians/groups/models.py
@@ -87,16 +87,23 @@ class GroupBase(models.Model):
         )
 
     def user_can_join(self, userprofile):
-        """Checks if a user can join a group."""
-        return (
-            # Must be vouched
-            userprofile.is_vouched and
-            # some groups don't allow
-            (getattr(self, 'accepting_new_members', 'yes') != 'no') and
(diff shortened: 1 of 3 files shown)


mozillians/groups/models.py
```diff
@@ -87,16 +87,23 @@ def user_can_leave(self, userprofile):
...
+            # Must be vouched and group must be Open
+            userprofile.is_vouched and getattr(self, 'accepting_new_members',
+                                               Group.OPEN) != Group.CLOSED,
+            # only makes sense to join if not already a member
+            not self.has_member(userprofile=userprofile),
+            # or the membership is in a pending state
+            not self.has_pending_member(userprofile=userprofile),
+            # If this is an access group the user should be able to join
+            (userprofile.can_join_access_groups()
+             if type(self) is Group and self.is_access_group else True)
```
johngian: Shouldn't this be ``else False`` ?

akatsoulas: I believe it should be `True`. This is basically like saying "If it's not an access group, don't affect the outcome of the `all` calculation.

PR #1994, opened 2018-01-12
Upgrade Celery to version 4.1.0

johngian:


mozillians/common/tasks.py
```diff
@@ -1,12 +1,12 @@
...
 
 import requests
-from celery.task import periodic_task
+
+from mozillians.celery import app
 
 
-@periodic_task(run_every=timedelta(hours=1))
+@app.task
+# (run_every=timedelta(hours=1))
```
akatsoulas: Let's remove this comment.

mozillians/celery.py
```diff
@@ -9,7 +9,36 @@
...
+    sender.add_periodic_task(RUN_DAILY, notify_membership_renewal.s(),
+                             name='notify-membership-renewal')
+
+    sender.add_periodic_task(RUN_DAILY, delete_reported_spam_accounts.s(),
+                             name='delete-reported-spam-accounts')
+
+    sender.add_periodic_task(RUN_HOURLY, celery_healthcheck.s(),
+                             name='celery-healthcheck')
+
+    sender.add_periodic_task(6 * RUN_HOURLY, periodically_send_cis_data.s(),
```
akatsoulas: Please add a comment here for the frequency of this task.

mozillians/celery.py
```diff
@@ -9,7 +9,36 @@
 
 from django.conf import settings  # noqa
 
+RUN_DAILY = 60 * 60 * 24
```
akatsoulas: Let's add configure Celery to log to Sentry. See the second example here:
https://docs.sentry.io/clients/python/integrations/celery/

mozillians/settings.py
```diff
@@ -322,15 +322,15 @@ def lazy_langs():
...
+CELERY_TASK_SERIALIZER = config('CELERY_TASK_SERIALIZER', default='pickle')
 CELERY_TASK_RESULT_EXPIRES = config('CELERY_TASK_RESULT_EXPIRES', default=3600, cast=int)
+CELERY_ENABLE_UTC = config('CELERY_ENABLE_UTC', default=True, cast=bool)
 CELERY_SEND_TASK_ERROR_EMAILS = config('CELERY_SEND_TASK_ERROR_EMAILS', default=True, cast=bool)
+CELERY_RESULT_BACKEND = config('CELERY_RESULT_BACKEND', default='amqp')
 REDIS_CONNECT_RETRY = config('REDIS_CONNECT_RETRY',
                              default=CELERY_RESULT_BACKEND == 'redis', cast=bool)
-# Time in seconds before celery.exceptions.SoftTimeLimitExceeded is raised.
-# The task can catch that and recover but should exit ASAP.
-CELERYD_TASK_SOFT_TIME_LIMIT = config('CELERYD_TASK_SOFT_TIME_LIMIT', default=150, cast=int)
```
akatsoulas: Probably we need to add the `CELERY_TRANSPORT_OPTIONS` configuration entry similar to this:
https://github.com/mozilla/remo/blob/master/remo/settings.py#L169

johngian: We are not scheduling ``eta`` tasks in mozillians and i would rather not add this workaround since we don't need it.

mozillians/settings.py
```diff
@@ -322,15 +322,15 @@ def lazy_langs():
 CELERY_ALWAYS_EAGER = config('CELERY_ALWAYS_EAGER', default='True', cast=bool)
 BROKER_CONNECTION_TIMEOUT = config('BROKER_CONNECTION_TIMEOUT', default=0.1, cast=float)
 CELERY_RESULT_BACKEND = config('CELERY_RESULT_BACKEND', default='amqp')
-BROKER_URL = config('BROKER_URL', default='amqp://guest:guest@broker:5672//')
-CELERY_ACCEPT_CONTENT = config('CELERY_ACCEPT_CONTENT', default='pickle', cast=Csv())
+CELERY_BROKER_URL = config('BROKER_URL', default='amqp://guest:guest@broker:5672//')
+CELERY_ACCEPT_CONTENT = ['pickle', 'json']
+CELERY_TASK_SERIALIZER = config('CELERY_TASK_SERIALIZER', default='pickle')
 CELERY_TASK_RESULT_EXPIRES = config('CELERY_TASK_RESULT_EXPIRES', default=3600, cast=int)
+CELERY_ENABLE_UTC = config('CELERY_ENABLE_UTC', default=True, cast=bool)
```
akatsoulas: You can drop this completely to simplify the config. It's `True` by default.

mozillians/settings.py
```diff
@@ -322,15 +322,15 @@ def lazy_langs():
 CELERY_ALWAYS_EAGER = config('CELERY_ALWAYS_EAGER', default='True', cast=bool)
 BROKER_CONNECTION_TIMEOUT = config('BROKER_CONNECTION_TIMEOUT', default=0.1, cast=float)
 CELERY_RESULT_BACKEND = config('CELERY_RESULT_BACKEND', default='amqp')
-BROKER_URL = config('BROKER_URL', default='amqp://guest:guest@broker:5672//')
-CELERY_ACCEPT_CONTENT = config('CELERY_ACCEPT_CONTENT', default='pickle', cast=Csv())
+CELERY_BROKER_URL = config('BROKER_URL', default='amqp://guest:guest@broker:5672//')
+CELERY_ACCEPT_CONTENT = ['pickle', 'json']
```
akatsoulas: Is there a reason why are we not going with `json` only (default)?

mozillians/settings.py
```diff
@@ -322,15 +322,15 @@ def lazy_langs():
 CELERY_ALWAYS_EAGER = config('CELERY_ALWAYS_EAGER', default='True', cast=bool)
 BROKER_CONNECTION_TIMEOUT = config('BROKER_CONNECTION_TIMEOUT', default=0.1, cast=float)
```
akatsoulas: Isn't this related to `amqp`? If yes, probably we can drop it?

mozillians/settings.py
```diff
@@ -322,15 +322,15 @@ def lazy_langs():
...
 CELERY_RESULT_BACKEND = config('CELERY_RESULT_BACKEND', default='amqp')
-BROKER_URL = config('BROKER_URL', default='amqp://guest:guest@broker:5672//')
-CELERY_ACCEPT_CONTENT = config('CELERY_ACCEPT_CONTENT', default='pickle', cast=Csv())
+CELERY_BROKER_URL = config('BROKER_URL', default='amqp://guest:guest@broker:5672//')
+CELERY_ACCEPT_CONTENT = ['pickle', 'json']
+CELERY_TASK_SERIALIZER = config('CELERY_TASK_SERIALIZER', default='pickle')
```
akatsoulas: Same here, why not `json`?

mozillians/settings.py
```diff
@@ -322,15 +322,15 @@ def lazy_langs():
...
 CELERY_TASK_RESULT_EXPIRES = config('CELERY_TASK_RESULT_EXPIRES', default=3600, cast=int)
+CELERY_ENABLE_UTC = config('CELERY_ENABLE_UTC', default=True, cast=bool)
 CELERY_SEND_TASK_ERROR_EMAILS = config('CELERY_SEND_TASK_ERROR_EMAILS', default=True, cast=bool)
+CELERY_RESULT_BACKEND = config('CELERY_RESULT_BACKEND', default='amqp')
 REDIS_CONNECT_RETRY = config('REDIS_CONNECT_RETRY',
```
akatsoulas: This is deprecated. Let's drop it.

mozillians/settings.py
johngian: A lot of stuff we are passing are not json serializable. (we were using pickle in the past by default)

mozillians/settings.py
johngian: Same here.

mozillians/settings.py
johngian: It still applies for redis.

mozillians/celery.py
johngian: Done a couple of lines below

PR #1995, opened 2018-01-15
Fix ADMINS config entry format.

johngian:


johngian:
@akatsoulas r?

This came up while skimming through the celery logs.

mozillians/settings.py
```diff
@@ -121,7 +121,7 @@
 DEBUG = config('DEBUG', default=False, cast=bool)
 
 ADMIN_ALIAS = config('ADMIN_ALIAS', default='mozillians-admins@mozilla.com')
-ADMINS = [('Mozillians.org Admins', ADMIN_ALIAS)],
+ADMINS = (('Mozillians.org Admins', ADMIN_ALIAS),)
```
akatsoulas: Why do we need to replace the list with a tuple?

johngian: It's from the documentation:
https://docs.djangoproject.com/en/1.8/ref/settings/#admins

Also the real issue is the trailing ``,``.

PR #2000, opened 2018-01-18
Remove the user from all the access groups if the user is removed fro…

akatsoulas:
…m NDA.

* Update the curators of an access group if a user removes the profile.

### DIFF
--- a/mozillians/groups/models.py
+++ b/mozillians/groups/models.py
@@ -86,6 +86,12 @@ class GroupBase(models.Model):
              self.has_pending_member(userprofile=userprofile))
         )
 
+    def curator_can_leave(self, userprofile):
+        """Check if a curator can leave a group."""
+        curators = self.curators.all()
+
+        return curators.count() > 1 or userprofile not in curators
+
     def user_can_join(self, userprofile):
         """Checks if a user can join a group.
 
@@ -353,6 +359,9 @@ class Group(GroupBase):
         If a user is a member of a reviewed or closed group,
         then the membership is in a pending state.
         """
+        # Avoid circular dependencies
+        from mozillians.users.models import UserProfile
+
         try:
             membership = GroupMembership.objects.get(group=self, userprofile=userprofile)
         except GroupMembership.DoesNotExist:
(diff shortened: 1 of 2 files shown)


mozillians/groups/models.py
```diff
@@ -367,10 +376,31 @@ def remove_member(self, userprofile, status=None, send_email=False):
...
+                    group_memberships = GroupMembership.objects.filter(userprofile=userprofile,
+                                                                       group__is_access_group=True)
+
+                for access_membership in group_memberships:
+                    if not access_membership.group.curator_can_leave(userprofile):
+                        # If the user is the only curator, let's add the superusers as curators
+                        # as a fallback option
+                        for super_user in UserProfile.objects.filter(user__is_superuser=True):
+                            access_membership.group.curators.add(super_user)
+                    access_membership.group.curators.remove(userprofile)
```
johngian: This way we might end up with curators that are not group members. Is that expected?

mozillians/groups/models.py
```diff
@@ -367,10 +376,31 @@ def remove_member(self, userprofile, status=None, send_email=False):
...
             # or the curator manually declined a user in a pending state.
             membership.delete()
-            send_userprofile_to_cis.delay(membership.userprofile.pk)
             # delete the invitation to the group if exists
             Invite.objects.filter(group=self, redeemer=userprofile).delete()
             send_email = True
+            # Remove all the access groups the user is a member of
+            # if the group to remove is the NDA
+            if self.name == settings.NDA_GROUP:
+                group_memberships = []
```
johngian: Instead of an empty list let's use an empty queryset for consistency.

``UserProfile.objects.none()``

mozillians/users/signals.py
```diff
@@ -26,6 +27,24 @@ def delete_user_obj_sig(sender, instance, **kwargs):
...
+
+    is a curator.
+    """
+    groups = Group.objects.filter(is_access_group=True, curators=instance)
+    for group in groups:
+        # If the user is the only curator of an access group
+        # add all the super users as curators and remove the user
+        if not group.curator_can_leave(instance):
+            for super_user in UserProfile.objects.filter(user__is_superuser=True):
+                group.curators.add(super_user)
```
johngian: Same here.

mozillians/groups/models.py
akatsoulas: Yes, the idea here is that administrators should find a solution for the case where there are no curators. It's just an administrative task, unrelated to the nature of the group that's why I didn't add them as members too. That said, I don't have a strong opinion on this.

PR #2002, opened 2018-01-22
Workaround for legacy PNGs and thumbnails. Convert RGBA to RGB.

johngian:


johngian:
Tested it locally:

* Manually added a legacy png to media files
* Saved path to obj
* Run ``get_photo_thumbnail``

master: Raises error regarding RGBA and JPEG
fix: Converts file and renders thumbnail correctly

johngian:
@akatsoulas r?

mozillians/users/models.py
```diff
@@ -518,7 +519,16 @@ def set_membership(self, model, membership_list):
     def get_photo_thumbnail(self, geometry='160x160', **kwargs):
         if 'crop' not in kwargs:
             kwargs['crop'] = 'center'
+
         if self.photo:
+            # Workaround for legacy images in RGBA model
+            image_obj = sorl_default.engine.get_image(self.photo)
+            if image_obj.mode == 'RGBA':
+                new_fh = default_storage.open(self.photo.name, 'w')
```
akatsoulas: Maybe use a context manager here?

johngian: Context manager in django storage backend was added in ``Django >= 2.0``

PR #2010, opened 2018-01-24
Remove patterns from Django urls.

akatsoulas:


johngian:
Code looks OK. I also did a diff between the master URL patterns and this branch and it returns the same number of URLs.

PR #2011, opened 2018-01-24
Send country information to basket.

johngian:


mozillians/users/tasks.py
```diff
@@ -120,12 +123,19 @@ def subscribe_user_task(self, result, email='', newsletters=[], sync='N', optin=
 
     if newsletters_to_subscribe:
         try:
+            kwargs = {
+                'sync': sync,
+                'source_url': MOZILLIANS_URL,
+                'optin': optin,
+                'api_key': BASKET_API_KEY
```
akatsoulas: Why not add both `email` and newsletters_to_subscribe` in kwargs too?

johngian: I am following  the signature definition from basket-client.

PR #2012, opened 2018-01-26
Remove duplicate distro package `node-less`.

tyler46:
I have just removed the duplicate entry of package `node-less` from both dockerfiiles.


tyler46:
happy to help :)

PR #2019, opened 2018-02-09
Split access groups and tags.

akatsoulas:
* Display the inviter to users that can create access groups.

### DIFF
--- a/mozillians/jinja2/phonebook/home.html
+++ b/mozillians/jinja2/phonebook/home.html
@@ -129,16 +129,33 @@
               </section>
             {% endif %}
             {% endwith %}
-            {% if user.userprofile.groups.count() %}
+            {% set access_groups = user.userprofile.get_annotated_access_groups() %}
+            {% if access_groups %}
               <section class="groups">
-                <h4>{{ _('Groups') }}</h4>
-                {% for group in user.userprofile.get_annotated_groups() -%}
+                <h4>{{ _('Access Groups') }}</h4>
+                {% for group in access_groups -%}
                   <a href="{{ url('groups:show_group', group.url) }}">
-                    {%- if user.userprofile in group.curators.all() -%}
+                    {%- if group.curators.filter(id=user.userprofile.id).exists() -%}
                       <i class="icon-crown"></i>
                     {% endif %}
                     {{ group.name }}
                     {%- if group.pending %} {{ _('(membership requested)') }}{% endif -%}
+                  </a>
+                  {%- if not loop.last -%},{% endif %}
+                {% endfor %}
+              </section>
+            {% endif %}
+            {% set tags = user.userprofile.get_annotated_tags() %}
+            {% if tags %}
+              <section class="groups">
+                <h4>{{ _('Tags') }}</h4>
+                {% for tag in tags -%}
+                  <a href="{{ url('groups:show_group', tag.url) }}">
+                    {%- if tag.curators.filter(id=user.userprofile.id).exists() -%}
+                      <i class="icon-crown"></i>
+                    {% endif %}
+                    {{ tag.name }}
+                    {%- if tag.pending %} {{ _('(membership requested)') }}{% endif -%}
                   </a>
                   {%- if not loop.last -%},{% endif %}
                 {% endfor %}
--- a/mozillians/jinja2/phonebook/profile.html
+++ b/mozillians/jinja2/phonebook/profile.html
@@ -220,10 +220,10 @@
           </div>
         {% endif %}
 
-        {% if groups %}
+        {% if access_groups %}
           <div id="groups" class="profile-entry">
-            <h3><i class="icon-group"></i> {{ _('Groups') }}</h3>
-              {% for group in groups %}
+            <h3><i class="icon-group"></i> {{ _('Access Groups') }}</h3>
+              {% for group in access_groups %}
                 {% set curators = group.curators.all() %}
                 {% if (user.is_authenticated() and user.userprofile.is_vouched) %}
                   <a href="{{ url('groups:show_group', group.url) }}">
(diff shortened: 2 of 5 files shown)


mozillians/jinja2/phonebook/home.html
```diff
@@ -129,9 +129,26 @@ <h2>{{ user.username }}</h2>
               </section>
             {% endif %}
             {% endwith %}
-            {% if user.userprofile.groups.count() %}
+            {% set access_groups = user.userprofile.get_annotated_groups(is_access_group=True) %}
+            {% if access_groups %}
```
johngian: Let's change this to ``if access_groups.exists()``.

mozillians/jinja2/phonebook/home.html
```diff
@@ -129,9 +129,26 @@ <h2>{{ user.username }}</h2>
...
               <section class="groups">
-                <h4>{{ _('Groups') }}</h4>
+                <h4>{{ _('Access Groups') }}</h4>
+                {% for group in access_groups -%}
+                  <a href="{{ url('groups:show_group', group.url) }}">
+                    {%- if user.userprofile in group.curators.all() -%}
```
johngian: Let's change this to something like: ``if group.curators.filter(<profile>=user.userprofile).exists()``
This way we are not iterating the qs on template rendering.

mozillians/users/models.py
```diff
@@ -623,28 +623,41 @@ def _email_now_vouched(self, vouched_by, description=''):
         send_mail(subject, filtered_message, settings.FROM_NOREPLY,
                   [self.email])
 
-    def get_annotated_groups(self):
+    def get_annotated_groups(self, is_access_group=False, get_inviter=False):
```
johngian: I don't agree with how you are changing this function.
You are adding logic about filtering in a function that just added an extra annotation. This is error prone because by default it returns only tags.

mozillians/jinja2/phonebook/profile.html
```diff
@@ -220,10 +220,43 @@ <h3><i class="icon-user"></i> {{ _('Bio') }}</h3>
...
+            <h3><i class="icon-group"></i> {{ _('Access Groups') }}</h3>
+              {% for group in access_groups %}
+                {% set curators = group.curators.all() %}
+                {% if (user.is_authenticated() and user.userprofile.is_vouched) %}
+                  <a href="{{ url('groups:show_group', group.url) }}">
+                    {%- if profile in curators -%}
+                      <i class="icon-crown"></i>
+                    {%- endif -%}
+                    {{ group.name }}
+                    {%- if group.pending %} {{ _('(membership requested)') }}{%- endif -%}
```
johngian: This way we might end up with the following text:
`` membership requested pending terms``

mozillians/users/models.py
```diff
@@ -623,28 +623,41 @@ def _email_now_vouched(self, vouched_by, description=''):
...
+            user_group_ids = list(groups_manager.values_list('id', flat=True))
+
+        for membership in self.groupmembership_set.filter(group__id__in=user_group_ids):
             group = membership.group
             group.pending = (membership.status == GroupMembership.PENDING)
             group.pending_terms = (membership.status == GroupMembership.PENDING_TERMS)
+
+            if get_inviter:
+                try:
+                    invite = Invite.objects.get(group=membership.group)
```
johngian: This is going to fetch all the invites for the group not only the one related to each membership.

mozillians/jinja2/phonebook/profile.html
```diff
@@ -220,10 +220,43 @@ <h3><i class="icon-user"></i> {{ _('Bio') }}</h3>
...
+                    {%- if group.pending_terms %} {{ _('(pending terms review)') }}{%- endif -%}
+                  </a>
+                  {%- if group.inviter and user.userprofile.can_create_access_groups %}
```
johngian: ``group.inviter`` is not per user.

mozillians/jinja2/phonebook/profile.html
```diff
@@ -220,10 +220,43 @@ <h3><i class="icon-user"></i> {{ _('Bio') }}</h3>
           </div>
         {% endif %}
 
-        {% if groups %}
+        {% if access_groups %}
```
johngian: ``access_groups.exists()``

mozillians/jinja2/phonebook/profile.html
```diff
@@ -220,10 +220,43 @@ <h3><i class="icon-user"></i> {{ _('Bio') }}</h3>
...
+                    <i class="icon-crown"></i>
+                  {%- endif -%}
+                  {{ group.name }}
+                {%- endif -%}
+                {% if not loop.last %},{% endif %}
+              {% endfor %}
+          </div>
+        {% endif %}
+
+        {% if tags %}
```
johngian: ``tags.exists()``

mozillians/jinja2/phonebook/profile.html
```diff
@@ -220,10 +220,43 @@ <h3><i class="icon-user"></i> {{ _('Bio') }}</h3>
...
+          <div id="groups" class="profile-entry">
+            <h3><i class="icon-group"></i> {{ _('Tags') }}</h3>
+              {% for group in tags %}
```
johngian: This is really confusing now that we decoupled ``groups`` and ``tags``.

mozillians/users/models.py
```diff
@@ -623,28 +623,41 @@ def _email_now_vouched(self, vouched_by, description=''):
...
+        groups_manager = self.groups.visible().filter(is_access_group=is_access_group)
         # checks to avoid AttributeError exception b/c self.groups may returns
         # EmptyQuerySet instead of the default manager due to privacy controls
+        user_group_ids = []
         if hasattr(groups_manager, 'visible'):
-            user_group_ids = list(groups_manager.visible().values_list('id', flat=True))
-        else:
-            user_group_ids = []
-        for membership in self.groupmembership_set.filter(group_id__in=user_group_ids):
+            user_group_ids = list(groups_manager.values_list('id', flat=True))
```
johngian: I am not sure that you still need this since you are filtering this above. Although the initial reason that it was built that way (i think) it was to distinct groups and skills.

mozillians/jinja2/phonebook/home.html
akatsoulas: The `access_groups` here is a list and not a queryset.

mozillians/jinja2/phonebook/profile.html
akatsoulas: Again this is a list.

mozillians/jinja2/phonebook/profile.html
akatsoulas: Same here.

mozillians/users/models.py
akatsoulas: You need to get all the groups that the user is a member of. After that you are using these IDs to loop over the memberships in order to get the `member/pending/pending terms` status.

PR #2024, opened 2018-02-20
Add parameter to disable silent auth.

johngian:


johngian:
@akatsoulas r?

mozillians/phonebook/views.py
```diff
@@ -591,6 +591,9 @@ def get(self, request):
             })
             request.session['oidc_verify_nonce'] = nonce
 
+        # Add parameter to disable silent authentication
+        params['tried_silent_auth'] = 'true'
```
akatsoulas: Let's add this as a var in the settings with a default value of 'true'.

PR #2026, opened 2018-02-27
Merge verification params into one.

akatsoulas:


mozillians/phonebook/views.py
```diff
@@ -591,11 +591,9 @@ def get(self, request):
             })
             request.session['oidc_verify_nonce'] = nonce
 
-        # Add parameter to disable silent authentication
-        params['tried_silent_auth'] = settings.OIDC_TRIED_SILENT_AUTH
-        # Add a parameter to disable the LDAP check for AUTO_VOUCH_DOMAINS
+        # Add parameter to disable silent authentication and the LDAP check for AUTO_VOUCH_DOMAINS
```
hidde: In NLX, I started using the word LDAP _shortcut_ check (it is the shortcut that lets users whose email passes the regex go without calling Person API to verify), we could consider saying “LDPA shortcut check” here?

akatsoulas: I prefer to explicitly describe the functionality that the param adds because there is nowhere in mozillians the term `shortcut check` and if you refer to this line of code in the future it will be rather difficult to recall what this means.

akatsoulas:
thanks @hidde!

PR #2033, opened 2018-03-05
bug 1442745 upgrade bleach to 2.1.3

g-k:
Upgrade bleach to 2.1.3 to fix a security issue escaping URIs
containing html entities affecting bleach 2.1 to 2.1.2. On mozillians
this is mitigated by CSP.


johngian:
Thanks for the patch @g-k

PR #2044, opened 2018-03-13
Remove stale log settings.

johngian:


johngian:
@akatsoulas r?

mozillians/settings.py
```diff
@@ -132,19 +131,6 @@
 # Site ID is used by Django's Sites framework.
 SITE_ID = 1
 
-# Log settings
-LOG_LEVEL = logging.INFO
-HAS_SYSLOG = config('HAS_SYSLOG', default=True, cast=bool)
-LOGGING_CONFIG = None
-
-SYSLOG_TAG = config('SYSLOG_TAG', default="http_app_mozillians")
-LOGGING = {
```
akatsoulas: This probably should be `LOGGING={}`

akatsoulas: Better yet let's update the `LOGGING` to something similar like this [1] and if we have the `DEBUG` flag to True we can change the handler to console.

[1] https://docs.djangoproject.com/en/1.11/topics/logging/#examples

mozillians/settings.py
```diff
@@ -661,25 +647,58 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
 
 
 LOGGING = {
```
akatsoulas: Move this above the `if DEV:` statement so that we can have all the dev and debug settings grouped together.

PR #2103, opened 2018-03-22
Call get_response after handling the locale in the request.

akatsoulas:
* Fixes the random locales in EXEMPT_URLS.


akatsoulas:
@fiji-flo r?

This fixes the random change of locales in the /admin interface. Basically, we  need to call the `get_response` after we modify to request to use the default locale.

PR #2143, opened 2018-04-03
Send all profiles to CIS in batches.

akatsoulas:


johngian:
Please double check that celery routing works.

akatsoulas:
I double checked with flower that the `cis` queue is used for this.

PR #2200, opened 2018-04-16
Initial orgchart implementation

johngian:


johngian:
@akatsoulas r?

mozillians/phonebook/utils.py
```diff
@@ -19,3 +26,57 @@ def redeem_invite(redeemer, code):
     invite.redeemer = redeemer
     invite.send_thanks()
     invite.save()
+
+
+def get_orgchart():
+    """Generate orgchart dict from s3 json dump."""
+
+    s3 = boto3.resource('s3')
+    orgchart_json = s3.Object(settings.ORGCHART_BUCKET, settings.ORGCHART_KEY).get()['Body'].read()
```
fiji-flo: Can we add an offline version to this? Either file or an S3 proxy.

mozillians/users/tests/__init__.py
```diff
@@ -86,3 +89,32 @@ class CityFactory(factory.DjangoModelFactory):
 
     class Meta:
         model = City
+
+
+class MockOrgChart:
+
+    @classmethod
+    def build(self, parent=None, depth=0, nodes=0):
+        """Recursively build mock orgchart."""
```
fiji-flo: We should generate a JSON structure like the one on S3 and parse that.

mozillians/phonebook/utils.py
```diff
@@ -19,3 +26,57 @@ def redeem_invite(redeemer, code):
     invite.redeemer = redeemer
     invite.send_thanks()
     invite.save()
+
+
+def get_orgchart():
```
akatsoulas: Probably you should rename this to something that better describes what this function does (eg create_orgchart).

mozillians/phonebook/views.py
```diff
@@ -743,3 +745,30 @@ def delete_idp_profiles(request):
     request.user.userprofile.idp_profiles.all().delete()
     messages.warning(request, 'Identities deleted.')
     return redirect('phonebook:profile_edit')
+
+
+@waffle_flag('view-orgchart')
+@never_cache
+def orgchart(request):
```
akatsoulas: Although the views are behind waffle, shouldn't we restrict them to only logged in or vouched users?

johngian: By default all views are restricted to logged-in/vouched

PR #2237, opened 2018-04-26
Avoid importing mock orgchart in prod.

johngian:


johngian:
@akatsoulas @fiji-flo r?

PR #2255, opened 2018-05-03
Start linking orgchart entries to staff profiles.

fiji-flo:
- basic correlaton via idp email matching
- some visual enchancements

### DIFF
--- a/mozillians/phonebook/utils.py
+++ b/mozillians/phonebook/utils.py
@@ -7,8 +7,10 @@ from anytree import Node
 from anytree.exporter import DictExporter
 
 from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist, MultipleObjectsReturned
 
 from mozillians.phonebook.models import Invite
+from mozillians.users.models import IdpProfile
 
 
 def redeem_invite(redeemer, code):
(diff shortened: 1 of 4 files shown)


mozillians/phonebook/utils.py
```diff
@@ -89,3 +92,11 @@ def create_orgchart():
 
     exporter = DictExporter()
     return exporter.export(nodes['root'])
+
+
+def get_profile_link_by_email(email):
+    try:
+        idp_profile = IdpProfile.objects.get(email=email)
+        return idp_profile.profile.get_absolute_url()
```
akatsoulas: Please move the return statement outside the `try/except` block. Ideally the try block should contain only the code that can raise the exception. Maybe you can use a `try/except/else` statement here.

mozillians/phonebook/utils.py
```diff
@@ -89,3 +92,11 @@ def create_orgchart():
...
+    except (ObjectDoesNotExist, MultipleObjectsReturned):
```
akatsoulas: You can use `IdpProfile.DoesNotExist` and `IdpProfile.MultipleObjectsReturned`

mozillians/phonebook/utils.py
fiji-flo: Thanks for the feedback.

mozillians/phonebook/utils.py
fiji-flo: done

fiji-flo:
Something broke the `test_send_renewal_notification` tests. Can anyone take a look? Otherwise, I'll debug this whenever I can make some time.

mozillians/phonebook/utils.py
```diff
@@ -89,3 +91,12 @@ def create_orgchart():
 
     exporter = DictExporter()
     return exporter.export(nodes['root'])
+
+
+def get_profile_link_by_email(email):
+    try:
+        idp_profile = IdpProfile.objects.get(email=email)
```
akatsoulas: let's add `primary=True` here too. There is only one email that can be primary whereas the same email can exist multiple times.

akatsoulas:
Could you please also squash before mergin?

PR #2268, opened 2018-05-08
Automatically delete akismet profiles that are non-vouched.

johngian:


mozillians/users/tasks.py
```diff
@@ -286,9 +286,8 @@ def delete_reported_spam_accounts():
     from mozillians.users.models import AbuseReport
 
     # Manual reports deletion heuristic:
-    # Delete all unvouched profiles reported as spam by a user
-    reports = AbuseReport.objects.filter(profile__is_vouched=False,
-                                         reporter__isnull=False)
+    # Delete all unvouched profiles reported
```
akatsoulas: Please update the comment to clearly state that we are deleting both akismet and user reported accounts. Also please update the task name now that you included akismet too.

mozillians/users/tasks.py
```diff
@@ -286,9 +286,8 @@ def delete_reported_spam_accounts():
...
+    reports = AbuseReport.objects.filter(profile__is_vouched=False)
     for report in reports:
```
akatsoulas: Shall we add a `try/except` here? I saw a couple of tasks failing because the profile returned a `DoesNotExist` exception.

PR #2280, opened 2018-05-10
Do not use link if there is not a profile.

akatsoulas:


mozillians/static/mozillians/js/orgchart.js
```diff
@@ -14,9 +14,12 @@ $(document).ready(function() {
       var div = document.createElement("div");
       var name = document.createTextNode(node.name);
 
-      link.setAttribute("href", node.href);
-
-
+      if (node.href) {
+        link = document.createElement("a");
```
johngian: remove the link reference above.

PR #2305, opened 2018-05-18
GitHub username

akatsoulas:


mozillians/jinja2/phonebook/includes/profile_edit_basic.html
```diff
@@ -127,6 +127,11 @@
                 <span class="identity-profile">
                   {{ form.instance.get_type_display() }} -
                   {{ form.instance.email }}
+                  {% if form.instance.username %}
+                    - <a href="https://github.com/{{ form.instance.username }}">
```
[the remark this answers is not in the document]
akatsoulas: Yes, this is basically a bad way of doing this because it won't work if we have usernames for other OPs. I will update this to be more OP agnostic.

mozillians/phonebook/views.py
```diff
@@ -582,7 +582,7 @@ def get(self, request):
 
         params = {
             'response_type': 'code',
-            'scope': import_from_settings('OIDC_RP_SCOPES', 'openid email'),
+            'scope': import_from_settings('OIDC_RP_SCOPES', 'openid email profile'),
```
akatsoulas: @fiji-flo thoughts about this? Basically in order to get the username we need to augment the scope by requesting data from the user profile.

mozillians/common/authbackend.py
```diff
@@ -111,9 +111,18 @@ def check_authentication_method(self, user):
             email=email,
             auth0_user_id=auth0_user_id)
 
+        # Update/Save the Github username
+        if 'github|' in auth0_user_id:
+            github_username = self.claims.get('nickname')
+            if current_idp:
+                current_idp.username = github_username
+                current_idp.save()
```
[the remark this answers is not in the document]
fiji-flo: Sorry, I'm not able to follow you there.

`current_idp = get_object_or_none(IdpProfile, profile=profile, primary=True)` retrieves the current active profile. Let's say this is _github.com/fiji-flo_. But now I'm adding a second profile that is linked to a different email address called _github.com/foobar_. Won't that set the `username` of both idps to `foobar`?

What about the last case?

mozillians/phonebook/views.py
fiji-flo: I guess that's just the way oidc works? :D

mozillians/common/authbackend.py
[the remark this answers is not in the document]
akatsoulas: Duh, it took me a while but I now totally get what you are saying. We don't need to update the current_idp never here. Only the obj which is queried against the auth0_id. This way we will always update the username in the correct profile.

mozillians/jinja2/phonebook/includes/profile_edit_basic.html
```diff
@@ -127,6 +127,10 @@
                 <span class="identity-profile">
                   {{ form.instance.get_type_display() }} -
                   {{ form.instance.email }}
+                  {% set username=form.instance|get_idp_external_url %}
+                  {% If username %}
```
johngian: Typo ``If``

mozillians/phonebook/templatetags/helpers.py
```diff
@@ -79,3 +80,19 @@ def get_search_models(models):
...
+    """Returns a link to the OP based on available options.
+
+    Returns a link to GitHub for obj.type == PROVIDER_GITHUB
+    Returns the username as a fallback for other OPs.
+    Returns an empty string if there is no username
+    """
+    if obj.username:
+        if obj.type == IdpProfile.PROVIDER_GITHUB:
+            github_url = '<a href="https://github.com/{username}" target="_blank">{username}</a>'
+            return jinja2.Markup(github_url.format(username=obj.username))
```
johngian: Should we escape this? By default its marked as safe string (i think)

mozillians/phonebook/views.py
```diff
@@ -691,6 +691,13 @@ def get(self, request):
                 'email': email
             }
 
+            # If we are linking GitHub we need to save
+            # the username too.
+            if 'github|' in user_info['sub']:
+                user_q.update({
```
johngian: Why not just do:
```
user_q['username'] = user_info['nickname']
```

akatsoulas: why not `update`?

mozillians/phonebook/templatetags/helpers.py
akatsoulas: I will escape the username here.

johngian:
Please fix failing tests.

PR #2324, opened 2018-05-24
Use privacy aware photos in profile page.

johngian:


johngian:
Some comments about the PR that were not clear in the first review:

* The initial reason that gravatars were not showing up is that we were using the "low-level" method that didn't fallback to gravatar

* That said there is a privacy issue that came up after the fix. The gravatar URL that we generate includes an md5 sum of the email. If the user is not exposing the email there is no way to generate a picture without exposing the email (via md5).

At the moment I added a check if the email is available at the point we generate the gravatar URL and if not it falls back the default avatar URL.

The problem is that even though we are trying to protect the privacy, it's not clear to the user what happens in the background and it might feel like a bug (eg. a user exposes the photo and shows the gravatar but for public users it returns no avatar).

@akatsoulas thoughts ? An other way to think about is that we expose md5 hashes and not the emails but still I am not that confident with the idea.

akatsoulas:
I would prefer not to expose the md5 hashes. Another, low tech, idea could be to have a check if there is a conflict (private email and public gravatar) and add a message in the profile settings to inform the user that the gravatar will not be exposed and propose a couple of solutions. Eg make the email public or upload a photo. Thoughts?

gene1wood:
> I would prefer not to expose the md5 hashes

Why's that? This is the standard way that every site that uses gravatar shows user avatars.

PR #2349, opened 2018-05-31
Allow users to select primary login account in Auth0.

akatsoulas:


akatsoulas:
@gdestuynder r?

mozillians/common/auth0.py
```diff
@@ -0,0 +1,38 @@
...
+    def get_access_token(self):
+        """Handle the access_token.
+
+        Re-use the existing token if it's valid else
+        get a new one.
+        """
+        token = self.client.access_token
+        if not token:
+            return self.client.get_access_token()
+        if self.client.access_token_valid_until < time.time():
```
gdestuynder: https://github.com/mozilla-iam/authzerolib/pull/7
either works though

mozillians/common/authbackend.py
```diff
@@ -119,7 +119,11 @@ def check_authentication_method(self, user):
         # Do not allow downgrades.
         # Round the current type to the floor.
         # This way 30 and 39 will have always the same priority.
-        if current_idp and obj.type < int(math.floor(current_idp.type / 10) * 10):
+        # If the type is less secure raise an error. If it is not but we have different
+        # raise again an error. Eg primary Firefox Accounts and the user tries to login
+        # with GitHub which has the same security level.
+        if (current_idp and (obj.type < int(math.floor(current_idp.type / 10) * 10) or
```
gdestuynder: nit: this seems more complicated than it should be

mozillians/common/auth0.py
```diff
@@ -0,0 +1,38 @@
...
+            'client_id': settings.OIDC_RP_CLIENT_ID,
+            'client_secret': settings.OIDC_RP_CLIENT_SECRET,
+            'uri': o.netloc
+        })
+
+    def change_primary_identiy(self, auth0_id, primary=False):
+        """Toggle status in the primary auth0 login identity."""
+        self.get_access_token()
+        user_schema = self.client.get_user(auth0_id)
+        app_metadata = user_schema['app_metadata']
```
gdestuynder: i suspect its possible that `app_metadata` is sometimes empty or not returned, i.e. you might want:
```
app_metadata = user_schema.get('app_metadata')
if app_metadata == None: app_metadata = {}
```
(or similar)

mozillians/common/auth0.py
```diff
@@ -0,0 +1,38 @@
...
+        app_metadata['mozilliansorg_primary'] = primary
```
gdestuynder: Just a note in case someone reads this later, we did not use the primary email attribute as it does not strongly tell you which is the login method (it just puts a name that contain the login method, which is an unverified text string)

hard to say which is better, and profile v2 will either need to strongly assert this or we'll need to actually implement account linking

mozillians/common/auth0.py
```diff
@@ -0,0 +1,38 @@
...
+        return self.client.update_user(auth0_id, app_metadata)
+
+    def get_access_token(self):
+        """Handle the access_token.
+
+        Re-use the existing token if it's valid else
+        get a new one.
+        """
+        token = self.client.access_token
+        if not token:
```
johngian: this can be moved to the if case.

```
if not token or self.client.valid:
    token = get_access_token()
return token
```

mozillians/common/authbackend.py
gdestuynder: this is how the auth0 rule works for example:
```
  var matchOrder = {'Mozilla-LDAP': 0,                                                                                                                                                                  
                    'Mozilla-LDAP-Dev': 0,                                                                                                                                                              
                    'firefoxaccounts': 1,                                                                                                                                                               
                    'github': 1,                                                                                                                                                                        
                    'google-oauth2': 2,                                                                                                                                                                 
                    'email': 3                                                                                                                                                                          
                   };
```
because its an object/dict multiple keys have the same value and then the code does
it means the subsequent check is slightly more verbose, but i believe that's actually a good thing (because you can read/understand instantly)

anyway, its just a nitpick

PR #2372, opened 2018-06-04
Check against the math.floor type of an Idp.

akatsoulas:


mozillians/users/models.py
```diff
@@ -740,7 +741,9 @@ def get_cis_groups(self, idp):
 
         # if the current idp does not match
         # the greatest number in the list, wipe the groups
-        if not idps or idp.type != max(idps) or not idp.is_mfa():
+        max_account_security_level = int(math.floor(max(idps) / 10) * 10)
```
johngian: nit: We can always create a method for that operation

PR #2375, opened 2018-06-05
Revert equally secure accounts.

akatsoulas:
* Firefox Accounts is more secure than GitHub.
* Re-enable the OIDC Refresh Token middleware.
* Display which account a user must use in order to login in the error message.


johngian:
Let's check if there are any stale data/schema migrations that need to be reverted.

PR #2455, opened 2018-07-02
Fix docker build

johngian:
* Fix broken dependency versions
* Pin pip < 10
* Change missing package name for libmysqlclient


johngian:
@akatsoulas r?

PR #2461, opened 2018-07-02
Do not allow editing `type` of access groups and tags after their cre…

akatsoulas:
…ation.

### DIFF
--- a/mozillians/groups/forms.py
+++ b/mozillians/groups/forms.py
@@ -303,37 +303,6 @@ class GroupCriteriaForm(happyforms.ModelForm):
         }
 
 
-class GroupAccessForm(happyforms.ModelForm):
-    """Modelform which handles the editing of access/tag groups."""
-
-    def __init__(self, *args, **kwargs):
-        self.request = kwargs.pop('request', None)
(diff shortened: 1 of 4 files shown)


mozillians/groups/forms.py
```diff
@@ -371,6 +341,12 @@ def clean(self):
         Check that only closed/reviewed groups can be access groups.
         """
         cdata = super(GroupCreateForm, self).clean()
+        user = getattr(self.request, 'user', None)
+        if ((not user or not user.userprofile.can_create_access_groups) and
+                self.cleaned_data['is_access_group']):
```
johngian: this shouldn't come from cleaned_data but from the stored instance.

mozillians/groups/forms.py
```diff
@@ -371,6 +341,12 @@ def clean(self):
         Check that only closed/reviewed groups can be access groups.
         """
         cdata = super(GroupCreateForm, self).clean()
+        user = getattr(self.request, 'user', None)
+        if ((not user or not user.userprofile.can_create_access_groups) and
```
johngian: if (not user) user.userprofile is going to fail.

PR #2466, opened 2018-07-03
Support privacy aware access to IDP emails.

johngian:
* Return privacy aware IDP when accessing profile.email
* Change UserProfile.privacy_email when saving primary identity

### DIFF
--- a/mozillians/users/models.py
+++ b/mozillians/users/models.py
@@ -334,7 +334,7 @@ class UserProfile(UserProfilePrivacyModel):
             # Try IDP contact first
             if self.idp_profiles.exists():
                 contact_ids = self.identity_profiles.filter(primary_contact_identity=True)
-                if contact_ids.exists():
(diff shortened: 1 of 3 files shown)


mozillians/users/models.py
```diff
@@ -861,6 +861,11 @@ def save(self, *args, **kwargs):
 
         super(IdpProfile, self).save(*args, **kwargs)
 
+        if self.primary_contact_identity:
```
akatsoulas: Please add a short comment on why this is here.

mozillians/jinja2/includes/search_result.html
```diff
@@ -67,7 +67,7 @@ <h2>
               </h2>
             </li>
           {% endif %}
-          {% if profile.email and privacy_level <= profile.privacy_email %}
```
akatsoulas: Why are you removing this? Now that you are setting the `privacy_email` to follow the privacy setting from the idp, you need to expose that only if the privacy_level of the user matches the one for the privacy_email.

PR #2471, opened 2018-07-04
Add data migration for email privacy field.

johngian:


mozillians/users/migrations/0036_auto_20180704_0634.py
```diff
@@ -0,0 +1,24 @@
...
+from django.db import migrations
+
+
+def migrate_privacy_email(apps, schema_editor):
+    UserProfile = apps.get_model('users', 'UserProfile')
+    IdpProfile = apps.get_model('users', 'IdpProfile')
+
+    for idp in IdpProfile.objects.filter(primary_contact_identity=True):
+        UserProfile.objects.filter(pk=idp.profile.pk).update(privacy_email=idp.privacy)
+
```
akatsoulas: Please also add a backwards method that does nothing so that we can go backwards if needed.

PR #2476, opened 2018-07-06
Do not set the privacy of the email attr from Idp.

akatsoulas:


mozillians/users/models.py
```diff
@@ -334,7 +334,7 @@ def _primary_email(self):
             # Try IDP contact first
             if self.idp_profiles.exists():
                 contact_ids = self.identity_profiles.filter(primary_contact_identity=True)
-                if contact_ids.exists() and contact_ids[0].privacy < self._privacy_level:
+                if contact_ids.exists() and contact_ids[0].privacy >= self._privacy_level:
```
johngian: Let's remove ``and contact_ids...`` its redundant in ``_identity_profiles``.

PR #2528, opened 2018-07-24
Remove prompt param in account linking trigger

hidde:
This parameter, although it follows OIDC standard, has recently started causing Firefox Accounts to error, as described in https://github.com/mozilla-iam/auth0-custom-lock/issues/182. This week the issue also started happening in PROD, hence my proposal to remove the parameter for now.

We're still passing the `account_linking=true` parameter, I have updated NLX DEV to respect that the same way as `prompt=select_account`. 

When releasing to PROD, we should release NLX to PROD first, then Mozillians.


akatsoulas:
Let's hold the merge until we are ready to release to prod.

PR #2529, opened 2018-07-24
First take on the GraphQL implementation.

akatsoulas:


mozillians/graphql/query.py
```diff
@@ -0,0 +1,31 @@
+import requests
+
```
johngian: no line, import with alphabetic order

mozillians/graphql/query.py
```diff
@@ -0,0 +1,31 @@
...
+        data = json2obj(resp)
+        # Query based on user_id
+        user_id = kwargs.get('userId')
+        if user_id:
+            for profile in data:
+                if profile['user_id']['value'] == user_id:
+                    return [profile]
+            return None
+
+        if isinstance(data, list):
```
johngian: Is there a case that this API returns a single object not in a list? This could be simplified to ``return data``.

mozillians/graphql/schema.py
```diff
@@ -0,0 +1,190 @@
+import graphene
+
+from mozillians.graphql.utils import parse_datetime_iso8601
+
+
+class Alg(graphene.Enum):
+    """V2 Schema Alg Enum object for Graphene."""
```
johngian: ``V2 Schema Alg object``

mozillians/graphql/schema.py
```diff
@@ -0,0 +1,190 @@
...
+class Typ(graphene.Enum):
+    """V2 Schema Typ object for Graphene."""
+
+    JWT = 'JWT'
+    PGP = 'PGP'
+
+
+class Classification(graphene.Enum):
+    """V2 Schema Classification object for Graphene."""
+    MOZILLA_CONFIDENTIAL = 'MOZILLA CONFIDENTIAL'
```
johngian: Add extra line.

mozillians/graphql/schema.py
```diff
@@ -0,0 +1,190 @@
...
+        return parse_datetime_iso8601(self.get('last_modified'))
+
+    def resolve_created(self, info, **kwargs):
+        """Resolver to return a datetime object."""
+        return parse_datetime_iso8601(self.get('created'))
+
+
+class StandardAttributeDatetime(graphene.ObjectType):
+    """V2 Schema StandardAttributeDatetime object for Graphene."""
+    value = graphene.DateTime()
```
johngian: New line please.

mozillians/graphql/schema.py
```diff
@@ -0,0 +1,190 @@
...
+        """Custom resolver for the Bugzilla Identity.
+
+        Extract the Bugzilla Identity from the profile v2 schema.
+        """
+        return self.get('bugzilla.mozilla.org')
+
+    def resolve_google(self, info, **kwargs):
+        """Custom resolver for the Bugzilla Identity.
+
+        Extract the Bugzilla Identity from the profile v2 schema.
```
johngian: Extract the google oauth identity.

mozillians/graphql/schema.py
```diff
@@ -0,0 +1,190 @@
...
+
+class Identities(graphene.ObjectType):
+    """V2 Schema Identities object for Graphene."""
+
+    values = graphene.Field(IdentitiesValues)
+    signature = graphene.Field(Signature)
+    metadata = graphene.Field(Metadata)
+
+    def resolve_values(self, info, **kwargs):
+        return self.get('values')
```
johngian: Isn't this the default resolver behavior.

mozillians/graphql/schema.py
```diff
@@ -0,0 +1,190 @@
...
+            return self['values']
+        values = self.get('values')
+        if values:
+            return values.items()
+        return None
+
+
+class AccessInformation(graphene.ObjectType):
+    """V2 Schema AccessInformation object for Graphene."""
+    ldap = graphene.Field(StandardAttributeValues)
```
johngian: New line please.

mozillians/graphql/schema.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class StandardAttributeValues(graphene.ObjectType):
+    """V2 Schema StandardAttributeValues object for Graphene."""
+
+    values = graphene.List(graphene.String)
+    metadata = graphene.Field(Metadata)
+    signature = graphene.Field(Signature)
+
+    def resolve_values(self, info, **kwargs):
```
johngian: Can you elaborate a little bit about this ?

mozillians/graphql/utils.py
```diff
@@ -0,0 +1,46 @@
+import json
+from aniso8601 import parse_datetime
+
+
+class ProfileFactory(dict):
+
```
johngian: Docstrings please.

mozillians/graphql/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        super(ProfileFactory, self).__init__()
+        for k, v in mapping.items():
+            self.__setitem__(k, v)
+
+    def __setitem__(self, key, value):
+        if isinstance(value, dict):
+            value = ProfileFactory(value)
+        super(ProfileFactory, self).__setitem__(key, value)
+
+    def __getattr__(self, item):
```
johngian: Why do you need to override this ?

mozillians/graphql/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        try:
+            return self.__getitem__(item)
+        except KeyError:
+            raise AttributeError(item)
+
+    __setattr__ = __setitem__
```
johngian: Can you provide some more information about this ?

mozillians/graphql/views.py
```diff
@@ -0,0 +1,15 @@
+from django.http import Http404
+from django.views.decorators.csrf import csrf_exempt
+
+import waffle
+from graphene_django.views import GraphQLView
+
+
+class MozilliansGraphQLView(GraphQLView):
+
```
johngian: Docstring please.

mozillians/graphql/views.py
```diff
@@ -0,0 +1,15 @@
...
+    @csrf_exempt
```
johngian: Can we fix this at this point? I think leaving this here is prone to release it without csrf.

mozillians/schema.py
```diff
@@ -0,0 +1,10 @@
+import graphene
+
+from mozillians.graphql import query
+
+
+class Query(query.Query, graphene.ObjectType):
```
johngian: This seems rather obscure, can you describe what we are doing here ?

mozillians/graphql/query.py
akatsoulas: In theory yes. While I was building this I worked both with single profiles and a few of them. I expect us to change this resolver significantly in the future. For now, I can simplify it and if needed we can re-change it.

mozillians/graphql/utils.py
akatsoulas: This will enable us to update the dictionary by assigning an attribute to the object.
For example
```python
foo = ProfileFactory()
print foo # {}
foo.x = 1
print(foo) # {'x': 1}
```
without that if you set `foo.x=1` the object foo will have the attribute 1 but it won't be in the dictionary

mozillians/graphql/schema.py
akatsoulas: Sure this. Most of the times this is going to return a dictionary of dictionaries. But we have also the case of `tags` which is basically a structure like:
```python
{'tags': ['tag1', 'tag2']}
```

PR #2629, opened 2018-08-24
Show report form only for non vouched profiles.

johngian:


johngian:
This came up after a discussion with to figure out if the way that the report form works is a bug or a feature. It looks like that showing it only for non-vouched people is more valuable at this point.

@akatsoulas r?

akatsoulas:
r+ for the code. I am curious though why do we need this change. With that the report button will not be shown for vouched users. This is fine for spam profiles because these users are already vouched and verified that are real but with this change we are also removing the ability to mark a profile as abusive. Can you elaborate a bit on that? cc @hmitsch

akatsoulas:
@johngian shall we merge this?

PR #2733, opened 2018-09-24
Various fixes for GraphQL

akatsoulas:
*

### DIFF
--- a/mozillians/settings.py
+++ b/mozillians/settings.py
@@ -621,9 +621,8 @@ CSP_REPORT_URI = config('CSP_REPORT_URI', default='/en-US/capture-csp-violation'
 CSP_DEFAULT_SRC = (
     "'self'",
     'https://cdn.mozillians.org',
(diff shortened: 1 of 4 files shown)


mozillians/settings.py
```diff
@@ -621,16 +621,16 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
 CSP_DEFAULT_SRC = (
     "'self'",
     'https://cdn.mozillians.org',
-    'https://www.google.com/recaptcha/',
```
johngian: Why are you removing this?

johngian:
r+ Lets merge it for dev purposes. In case we need to expose graphiql to the public let's revisit and use the upstream fix when it gets landed.

mozillians/settings.py
akatsoulas: because it's in the script-src which is needed. No need to have it in the default policy.

PR #2743, opened 2018-09-26
k8s setup for dev preview

johngian:


johngian:
@fiji-flo r?

I haven't tested using the ingress controller but right now all the containers start without errors. Also we should decide how we spin off the shared services like: mysql, es, redis, ingress

johngian:
@fiji-flo r?

I haven't tested using the ingress controller but right now all the containers start without errors. Also we should decide how we spin off the shared services like: mysql, es, redis, ingress

PR #2805, opened 2018-10-11
adds buildspec.yml for CodeBuild

danielhartnell:
Request to add a basic `buildspec.yml` for CodeBuild. Once this is merged, I can configure CodeBuild to watch this repository for changes.

Let me know if I missed anything from the contribution guidelines. This will buildspec will tag each image with the short commit SHA when it pushes to ECR. With the current CodeBuild configuration I'm tracking stage builds on the `master` branch and `production` builds on the production branch. I'm not currently tracking the `cistesting` branch but I can add that when we need it.


johngian:
I have pushed a branch [1] with this change. Let's use that branch for the migration work.

[1] https://github.com/mozilla/mozillians/tree/iam-infra-migration

danielhartnell:
That's perfect. Thanks.

PR #2830, opened 2018-10-16
Route search/orgchart/graphql via mozillians.org

johngian:


mozillians/dino_park/views.py
```diff
@@ -1,10 +1,69 @@
...
+
+@never_cache
+@allow_public
+def orgchart(request):
+    """Internal routing to expose orgchart service."""
+    url_parts = urlparse.ParseResult(
+        scheme='http',
+        netloc=settings.DINO_PARK_ORGCHART_SVC,
+        path='/orgchart',
+        params='',
```
akatsoulas: You need a scope here too. Basically it's always the 'staff' scope

mozillians/dino_park/views.py
```diff
@@ -1,10 +1,69 @@
...
+from mozillians.dino_park.utils import UserAccessLevel
 
 
 @never_cache
 @allow_public
 def main(request):
     return render(request, 'dino_park/index.html', {})
+
+@never_cache
+@allow_public
```
akatsoulas: nope :)

mozillians/dino_park/views.py
```diff
@@ -1,10 +1,69 @@
...
+        fragment=''
+    )
+    url = urlparse.urlunparse(url_parts)
+    resp = requests.get(url)
+    resp.raise_for_status()
+    return JsonResponse(resp.json)
+
+@never_cache
+@allow_public
+def search_get(request, user_id):
```
akatsoulas: Probably `get_profile` is more relevant name.

johngian: This is following the service convention. I suggest we leave it the same to avoid confusion.

mozillians/dino_park/views.py
johngian: No this is not relevant anymore.

akatsoulas: Better said, we need to only allow staff members to view the orgchart (reach this view)

PR #2850, opened 2018-10-18
Iam infra migration

danielhartnell:
Rebases against `mozillians/dino-park` and adds support for S3 media backend.


danielhartnell:
I would recommend that we close this PR. The following two commits (shown above) are the only two that I want to retain. The rest were an attempt to keep this in sync with dino-park while I built and deployed the Docker image to Kubernetes.

I'll create a separate PR to add those commits to the dino-park branch so we can continue this work in one place. This will let me connect the CodeBuild job to that branch to build and deploy updates to the dino-park branch.

- adds buildspec.yml for CodeBuild
- add support for S3 media backend in prod 

I'll refer to the new PR once I create it and close this one.

PR #2889, opened 2018-10-26
Pin flake8 to verion 3.5.0.

akatsoulas:


akatsoulas:
This fixes failing tests due to flake upgrade to version 3.6.0

PR #2893, opened 2018-10-26
Pep8 changes

akatsoulas:


mozillians/urls.py
```diff
@@ -26,9 +26,9 @@ def error_page(request, template, status=None):
     return render(request, '%d.html' % template, status=(status or template))
 
 
-handler404 = lambda r: error_page(r, 404)
-handler500 = lambda r: error_page(r, 500)
-handler_csrf = lambda r, cb=None: error_page(r, 'csrf_error', status=400)
+handler404 = lambda r: error_page(r, 404) # noqa
```
johngian: you can always convert them to functions to avoid ``noqa``.

mozillians/users/api/v2.py
```diff
@@ -143,7 +143,7 @@ def transform_timezone(self, obj, value):
     def transform_bio(self, obj, value):
         return {
             'value': value,
-            'html': unicode(markdown(value)),
+            'html': unicode(markdown(value)), # noqa
```
johngian: why do we need ``noqa`` here?

akatsoulas: it complains about `unicode`

PR #2915, opened 2018-10-31
Get the user from request if not id is supplied.

akatsoulas:


akatsoulas:
Fixes #2905

akatsoulas:
Tests are failing due to flake8 upgrade. This is fixed in master.

PR #2925, opened 2018-11-01
add if block for resources that are not needed in staging and prod

danielhartnell:
Hey,

Sorry, the commit is a little hard to read. I grouped together the services and deployments that are not needed in the staging and prod environments and then wrapped them in a Go template if block. If we're using `dev.yaml` in the Helm template rendering, all resources will be created. If we're using `staging.yaml` or `prod.yaml`, MySQL, Redis, Memcached and ES should be excluded so we can use the AWS resources.

Let me know if you have any questions.


k8s/templates/deployment.yaml
```diff
@@ -35,128 +35,129 @@ spec:
 apiVersion: apps/v1
```
akatsoulas: Probably we need to add another if here[0] to use either the `run-prod.sh` or the `run-dev-preview.sh` depending on the environment.

[0] https://github.com/mozilla/mozillians/pull/2925/files#diff-b7fdde1eb4ef161de899591e53746444R29

PR #2934, opened 2018-11-05
removes hard-coded namespace in configmap

danielhartnell:
Hey,

This is currently resulting in failed deployments in Kubernetes (dinopark-dev namespace). Once this is merged, I can trigger a build to update the running pods.


akatsoulas:
reverting this, since it's handled in #2930

PR #2936, opened 2018-11-05
Temporarily disable W504 in flake.

akatsoulas:


akatsoulas:
Temporarily disables a couple of checks in flake8 to fix the failing tests in dino-park branch. This will be reverted when merged to master.

PR #2938, opened 2018-11-05
Fix deployment template

akatsoulas:


akatsoulas:
Closing this. The fix was pushed in dino-park branch.

PR #2960, opened 2018-11-06
Fix celery k8s.

fiji-flo:


k8s/templates/01-dependencies.yaml
```diff
@@ -24,13 +24,32 @@ spec:
...
+          {{ if eq .Values.env "staging" -}}
+          volumeMounts:
+            - mountPath: /etc/secrets-volume/
+              name: secrets-volume
+          {{- end }}
+      {{ if eq .Values.env "staging" -}}
+      volumes:
+        - name: secrets-volume
+          secret:
+            secretName: mozillians-dev
```
akatsoulas: nit: Maybe we can use a var here and everywhere else we are using this.

PR #2991, opened 2018-11-13
add Kubernetes requirements to the Helm templates

danielhartnell:
Hi,

Sorry, there are a few changes in this PR which are not dependent on each other. If you would prefer, I can break them out into separate pull requests or squash commits. Just let me know.

**New stuff:**

* The addition of `bin/run-k8s.sh`. The command to run on container initialization was getting pretty long and I did not want to add to it. Mozillians uses the CIS library which uses Everett and it looks for some `CIS_*` envars (seen [here](https://github.com/mozilla-iam/cis/blob/master/cis/settings.py)). In order to expose these, I just re-implemented something similar to what `bin/bootstrap.sh` was doing by exporting the values found in `.env`.

* A new annotation if the environment is staging which will allow me to specify a role for the pod to assume via kube2iam.

* A new `volumeMounts` config for the secrets which was referenced but unavailable previously.

I did struggle a little with the Go template newline settings. I had to play around with the hyphens to get the staging and dev templates to render correctly. It's possible I might have missed something so I would love a second set of eyes on that if you can.

PR #3004, opened 2018-11-13
add Kubernetes requirements to the Helm templates [duplicate of #2991]

danielhartnell:
This is a follow up to #2991 which brings the changes into master rather than `dino-park`.


danielhartnell:
Ignore this PR. I will open up a new one.

PR #3006, opened 2018-11-13
exposes configurable annotation for kube2iam

danielhartnell:
I decided to break #2991 into separate pull requests. This allows me to configure a role for the pod to assume with kube2iam in each pod.

### DIFF
--- a/k8s/templates/01-dependencies.yaml
+++ b/k8s/templates/01-dependencies.yaml
@@ -15,6 +15,10 @@ spec:
     metadata:
       labels:
         app: {{ .Values.name }}-celery
+      {{ if eq .Values.env "staging" -}}
+      annotations:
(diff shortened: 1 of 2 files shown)


k8s/templates/01-dependencies.yaml
```diff
@@ -15,6 +15,10 @@ spec:
     metadata:
       labels:
         app: {{ .Values.name }}-celery
+      {{ if eq .Values.env "staging" -}}
```
akatsoulas: maybe it makes sens to remove the hard-coded `staging` value and replace it with `ne dev`. @fiji-flo thoughts?

danielhartnell: This would help when we add a prod deployment as well. I'm open to the idea.

PR #3011, opened 2018-11-14
Enable s3 storage for dinopark deployment.

johngian:


johngian:
@akatsoulas r?

akatsoulas:
Can we replace this with pr #3001 ? we need the flag for other services too so maybe it makes sens to have only one

johngian:
Closing this one after #3001

PR #3023, opened 2018-11-15
Inline mozilliansorg username.

fiji-flo:


mozillians/graphql_profiles/resolvers.py
```diff
@@ -8,6 +8,11 @@
 def dino_park_resolver(attname, default_value, root, info, *args):
     """Custom resolver for all the attributes in a profile."""
 
+    # inline mozilliansorg username
+    if attname == 'username':
```
akatsoulas: Let's move this a bit below under `if profile_attr and hasattr(profile_attr, 'get'):` where everything related to the v2 profile are located.

https://github.com/mozilla/mozillians/pull/3023/files#diff-66ffe59432d122809127cb1e2588fde1R32

PR #3024, opened 2018-11-15
Enable seamless login for dinopark.

johngian:


mozillians/common/middleware.py
```diff
@@ -173,3 +174,15 @@ def __call__(self, request):
...
+
+
+class DinoParkLoginMiddleware(object):
+    """Seamless login for dinopark behind OIDC proxy"""
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if not request.user.is_authenticated() and not request.path.startswith('/oidc/'):
```
akatsoulas: Let's add all the exempt urls here.

PR #3048, opened 2018-11-19
Pass query to search.

fiji-flo:


mozillians/dino_park/views.py
```diff
@@ -69,7 +69,7 @@ def search_simple(request, query):
         netloc=settings.DINO_PARK_SEARCH_SVC,
         path='/search/simple/{}/{}'.format(scope, query),
         params='',
-        query='',
+        query=request.query,
```
johngian: It should be something like ``request.GET.urlencode()``

johngian: Does it also allow ``POST``? If yes you should handle it accordingly.

fiji-flo: Only get for now. Thanks.

PR #3052, opened 2018-11-20
Add oidc access group

danielhartnell:
Adding the group reference as an environment variable sourced from the Kubernetes secret. I also noticed a duplicate discovery_url secret so I removed it.

### DIFF
--- a/k8s/templates/oidc-proxy.yaml
+++ b/k8s/templates/oidc-proxy.yaml
@@ -51,6 +51,11 @@ spec:
                 secretKeyRef:
                   name: oidc-proxy-secret
                   key: redirect_uri
+            - name: allowed_group
+              valueFrom:
+                secretKeyRef:
+                  key: allowed_group
+                  name: oidc-proxy-secret
 
 ---
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


k8s/templates/oidc-proxy.yaml
```diff
@@ -36,11 +36,6 @@ spec:
                 secretKeyRef:
                   name: oidc-proxy-secret
                   key: discovery_url
-            - name: discovery_url
```
fiji-flo: This is still set in the currently running oidc proxy, are you sure we don't need it?

danielhartnell: @fiji-flo I'm sure we do need it. It's just duplicated so I removed one copy:

https://github.com/mozilla/mozillians/blob/ee955dca61350c0dd2ca4fc23c22dcf4083a79bd/k8s/templates/oidc-proxy.yaml#L34-L43

akatsoulas:
I am closing this PR since the OIDC access group is already added. @danielhartnell if there is a dup entry, could you please open a new PR for that?

PR #3065, opened 2018-11-21
Separate mozillians.org urls from DinoPark.

akatsoulas:


akatsoulas:
We are going to use a middleware to redirect everything to dino-park. closing this.

PR #3066, opened 2018-11-21
Force redirection to /beta for all user with complete profiles.

johngian:


johngian:
@akatsoulas r?

mozillians/common/middleware.py
```diff
@@ -191,4 +191,7 @@ def __call__(self, request):
         if not request.user.is_authenticated():
             return HttpResponseRedirect(django_reverse('oidc_authentication_init'))
 
+        if request.user.userprofile and request.user.userprofile.is_complete:
```
akatsoulas: You can just add an else clause to lines above. Also the `useprofile.is_complete` is not needed because the oidc flow will handle that by populating the profile if it does not exist.

PR #3081, opened 2018-11-26
Do not search orgchart for contributors.

akatsoulas:


mozillians/dino_park/views.py
```diff
@@ -47,6 +47,12 @@ def orgchart_get_by_username(request, path, username):
         user = User.objects.get(username=username)
     except User.DoesNotExist:
         user = None
+
+    # if there is a user and the user is not a staff member
+    # then we don't need to search for a profile in orgchart
+    if user and not user.userprofile.is_staff and path == 'trace':
+        return JsonResponse(None, safe=False)
```
johngian: why `safe=False` ? Shouldn't it be handles by `JsonResponse` ?

mozillians/dino_park/views.py
```diff
@@ -47,6 +47,12 @@ def orgchart_get_by_username(request, path, username):
         user = User.objects.get(username=username)
     except User.DoesNotExist:
         user = None
+
+    # if there is a user and the user is not a staff member
+    # then we don't need to search for a profile in orgchart
+    if user and not user.userprofile.is_staff and path == 'trace':
```
johngian: i am not sure but maybe ``trace`` should be ``/trace``

mozillians/dino_park/views.py
```diff
@@ -61,7 +67,6 @@ def orgchart_get_by_username(request, path, username):
     )
     url = urlparse.urlunparse(url_parts)
     resp = requests.get(url)
-    resp.raise_for_status()
```
johngian: what happens if response is not ``20x`` or ``30x`` ?

mozillians/dino_park/views.py
```diff
@@ -104,5 +109,4 @@ def search_get_profile(request, username, scope=None):
...
```
johngian: Same here

mozillians/dino_park/views.py
akatsoulas: These errors should eventually be handled in the frontend

mozillians/dino_park/views.py
akatsoulas: Without that the `data` payload should be only dictionary.

mozillians/dino_park/views.py
akatsoulas: `trace` should be fine. Slashes are added during the url construction

PR #3114, opened 2018-11-30
Add OpenSearch desc file for DinoPark.

akatsoulas:


akatsoulas:
Merging this as it was approved as part of #3113

PR #3163, opened 2018-12-17
Revert "Update settings for the verification Auth0 client."

akatsoulas:
Reverts mozilla/mozillians#3149


akatsoulas:
This is no longer needed since we are adding a new claim in the assertions.

PR #3164, opened 2018-12-17
Update verification client to get the original user id.

akatsoulas:


akatsoulas:
See also https://github.com/mozilla-iam/auth0-deploy/pull/246

PR #3427, opened 2019-03-13
Add mysqlclient to prod requirements.

akatsoulas:
* Add missing lib in dev requirements.


akatsoulas:
@johngian building against mariadb is failing with MySQL-Python.

johngian:
Code looks OK. Why do we need to build against mariadb though?

akatsoulas:
This has its fair share of problems with local dev if you are not using docker and also the package is replacing it's not maintained for a long time. It looks like it's just a simple swap. Shall we replace it? Thoughts?

PR #3434, opened 2019-03-15
Separate DinoPark from mozillians.org.

akatsoulas:


akatsoulas:
This is the first attempt to remove DinoPark from mozillians.org. I will go through the codebase in case there are more traces of DinoPark.

akatsoulas:
@fiji-flo just an fyi, although this is not ready yet to be merged and deployed.

PR #3708, opened 2019-06-27
Changed restart policy and added container names

kmehant:
I have changed the restart policy to `unless-stopped` which is better than `always`. Added container names as it makes simpler for attaching to running containers.


kmehant:
@glogiotatidis @akasantony Please review!

Issue #3727, opened 2019-07-17
Branding does not match the current Mozilla branding


WesleyBranton:
This is one of the many Mozilla websites that has not yet been updated to the current Mozilla brand image.

oluebubeoti:
[https://dribbble.com/shots/9705055-Mozillians-Logo-Redesign?utm_source=Clipboard_Shot&utm_campaign=oluebubeoti&utm_content=Mozillians%20Logo%20Redesign&utm_medium=Social_Share
![Mozillians Redesign](https://user-images.githubusercontent.com/52883655/72876404-b684cf80-3cf6-11ea-9db4-752d200f9473.png)
](url)

chasestruck:
Is this official? Why has it not been added yet?
If I can get confirmation that this is official and that there are plans to update the branding I could submit a PR with this new images (assuming that one is not already in the works)

chasestruck:
I'm sure everyone knows this by now, but it seems like mozillans.org is getting removed in favor of people.mozilla.org and community.mozilla,org hence why this is not getting reviewed

See: https://old.reddit.com/r/firefox/comments/f1tdee/pr_to_update_branding_on_mozillansorg_has_been_up/

WesleyBranton:
> I'm sure everyone knows this by now, but it seems like mozillans.org is getting removed in favor of people.mozilla.org and community.mozilla,org hence why this is not getting reviewed

It's strange because I've heard the opposite. As of October 2019, the plan was to keep Mozillians: **https://discourse.mozilla.org/t/frequently-asked-questions-portal-edition-faq/43224/8**

chasestruck:
Hm, that is strange. I can't imagine they would stall this PR for so long if the plan was to keep it, though. I think the rebrand was in like 2017? That would be almost 3 years on a trival logo fix...

oluebubeoti:
![custom – 1](https://user-images.githubusercontent.com/52883655/76142522-eaca0a80-606e-11ea-8298-c4906fa127d2.png)
Mozillians Website Redesign. View on dribbble https://dribbble.com/shots/10628036-Mozillians-Website-Redesign

PR #3728, opened 2019-07-17
Mozilla branding update

WesleyBranton:
This will fix bug #3727 which needs to be addressed.


ChiefMikeK:
Also don't forget to add the 
 moz://la logo to the Readme.md  :neckbeard:

WesleyBranton:
> 
> 
> Also don't forget to add the
> moz://la logo to the Readme.md :neckbeard:

The logo wasn't on the readme to begin with, so why would it need to be added now?

PR #3836, opened 2019-08-21
Add a second NDA group.

akatsoulas:


mozillians/settings.py
```diff
@@ -280,6 +280,9 @@ def lazy_langs():
 
 # NDA Group
 NDA_GROUP = config('NDA_GROUP', default='nda')
+# TODO change this
+NDA_STAFF_GROUP = config('NDA_STAFF_GROUP', default='')
```
johngian: i suggest you add a placeholder for now, to avoid config changes which might be tricky at this point.

mozillians/users/admin.py
```diff
@@ -268,6 +268,33 @@ def queryset(self, request, queryset):
         return queryset
 
 
+class NDAStaffMemberFilter(SimpleListFilter):
+    """Admin filter for profiles member of the NDA group"""
```
johngian: staff NDA group

mozillians/users/models.py
```diff
@@ -443,10 +443,12 @@ def is_manager(self):
     def is_nda(self):
         query = {
             'userprofile__pk': self.pk,
-            'group__name': settings.NDA_GROUP,
             'status': GroupMembership.MEMBER
         }
-        return GroupMembership.objects.filter(**query).exists() or self.user.is_superuser
+        is_nda_member = (GroupMembership.objects.filter(Q(group__name=settings.NDA_GROUP)
```
johngian: you could change that to `GroupMembership.objects.filter(group__name__in=[])` to avoid the extra complexity

mozillians/users/views.py
```diff
@@ -162,9 +162,10 @@ def get_queryset(self):
 
         # Query NDA memberships
         nda_members_ids = GroupMembership.objects.filter(
-            group__name=settings.NDA_GROUP,
-            status=GroupMembership.MEMBER
-        ).values_list('userprofile__pk', flat=True)
+            Q(group_name=settings.NDA_GROUP) | Q(group_name=settings.NDA_STAFF_GROUP)
```
johngian: same here

mozillians/settings.py
akatsoulas: Agreed but I wasn't sure what that should be. I ll think of something

PR #4016, opened 2019-10-27
fixed typo

divyamoncy:


maheshm:
Usually the commit messages should be "This commit will <your commit message>". So your commit message should be "fix typo". It could be a bit more descriptive as well.

